PR #7, opened 2018-04-03
add ref local reassignment

BillWagner:
And build project around the sample.

Sample updates for dotnet/docs#3961


snippets/csharp/programming-guide/ref-returns/NumberStore.cs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        public override string ToString()
+        {
+            string retval = "";
+            for (int ctr = 0; ctr < numbers.Length; ctr++)
+            {
+                retval += $"{numbers[ctr]} ";
+            }
+            return retval.Trim();
```
pkulikov: I see that is pre-existing code, but to make snippet shorter what about such `ToString()` implementation:
```csharp
public override string ToString() => string.Join(" ", numbers);
```

Same remark also for the *NumberStoreUpdated.cs* file

BillWagner:
This PR must not be merged until dotnet/docs#4875

BillWagner:
Thanks @rpetrusha That comment was a test and a reminder to add a [note](https://github.com/dotnet/docs/issues/4154#issuecomment-380243443) on another issue.

PR #9, opened 2018-04-07
Fix HttpClientHandler sample spacing

karelz:
Fix HttpClientHandler sample spacing

Primary reason: Attempt to fix weird 'space' before `static async void Main()` in the [HttpClientHandler sample](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler?view=netcore-2.0).
Not sure if it is going to help though.

### DIFF
--- a/snippets/csharp/VS_Snippets_Misc/system.net.http.httpclienthandler/cs/source.cs
+++ b/snippets/csharp/VS_Snippets_Misc/system.net.http.httpclienthandler/cs/source.cs
@@ -7,7 +7,7 @@ class HttpClientHandler_Example
 {
 // <Snippet1>
    static async void Main()
-	 {
+   {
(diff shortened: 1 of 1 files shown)


karelz:
BTW: I found similar problem in [HttpClient sample](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=netcore-2.0#examples) - [this file](https://github.com/dotnet/samples/blob/master/snippets/csharp/VS_Snippets_Misc/system.net.http.httpclient/cs/source.cs).
Happy to fix it too if the changes are expected to help with the "primary problem".

karelz:
Thanks @BillWagner!
BTW: I forked the repo and plan to submit larger spacing / formatting change to more samples as I discover them (while reading networking docs): https://github.com/karelz/samples/pull/1

PR #12, opened 2018-04-11
Minor formatting and readability improvements

Mackiovello:
* Consistently use Allman braces
* Change name of variable to be more descriptive
* Use null coalescing operator where appropriate
* Remove comment that didn't provide any value


Mackiovello:
I appreciate the detailed feedback @rpetrusha. I'll keep in mind to also update the VB.NET sample as well in the future.

PR #13, opened 2018-04-11
Update TryParse sample for readability

Mackiovello:
Use an array instead of repeating code.
Change variable name from "result" to "success" to be more semantic
Use null coalescing operator instead of null check


Mackiovello:
Thank you @rpetrusha for the feedback. I'll fix the things you mentioned and update the VB.NET example later today.

Mackiovello:
They should be consistent now, please check

PR #32, opened 2018-04-27
GenericWhereConstraints.cs: fix minor inconsistencies

pkulikov:
The [topic on where keyword](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/where-generic-type-constraint) says:
>For example, you can declare a generic class, `MyGenericClass`, such that the type parameter `T` implements the `IComparable<T>` interface:

I've fixed the code snippet to match that description. After that, I've also checked other places where generic interfaces might be used.

### DIFF
--- a/snippets/csharp/keywords/GenericWhereConstraints.cs
+++ b/snippets/csharp/keywords/GenericWhereConstraints.cs
@@ -8,7 +8,7 @@ using static keywords.UnmanagedExtensions;
 namespace keywords
 {
     // <Snippet1>
-    public class AGenericClass<T> where T : IComparable { }
+    public class AGenericClass<T> where T : IComparable<T> { }
     // </Snippet1>
 
     // <SNippet2>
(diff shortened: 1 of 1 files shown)


snippets/csharp/keywords/GenericWhereConstraints.cs
```diff
@@ -46,15 +46,13 @@ public interface IMyInterface { }
     namespace CodeExample
     {
         class Dictionary<TKey, TVal>
-            where TKey : IComparable, IEnumerable
+            where TKey : IComparable<TKey>
```
BillWagner: I'm curious why you dropped `IEnumerable`?

pkulikov: My confusion was: Why to enumerate through a key in the dictionary context? Quite a specific use case. But yes, of course, that might be. But then I was a little bit against non-generic interface in the example and `TKey : IEnumerable<TKey>` looked even more fancier. Thus, I gave up.

I would rather expect `class Dictionary<TKey, TVal> : IEnumerable<TVal>`, by the way.
If we want to show several interfaces in one `where` constraint, what about
```csharp
where TKey : IComparable<T>, IEquatable<T>
```
Or it's better not to mix those two interfaces without care?

BillWagner: Thanks for the extra explanation.

Removing `IEnumerable` is the right choice. It isn't enlightening the discussion, and is not likely a good constraint on the key type.

I'll :shipit: now.

PR #38, opened 2018-05-03
Csharp 73 autoproperty field attributes

BillWagner:
Sample code for dotnet/docs#3962


csharp/serialization/Loan.cs
```diff
@@ -0,0 +1,54 @@
...
+        public int Term { get; set; }
+
+        private string customer;
+        public string Customer
+        {
+            get { return customer; }
+            set
+            {
+                customer = value;
+                PropertyChanged(this,
```
pkulikov: @BillWagner shouldn't this line be
```csharp
PropertyChanged?.Invoke(this,
```

pkulikov: Oops, I've realized I can update it myself as it's already merged into the master

PR #43, opened 2018-05-04
Corrected snippets for Equals

mikkelbu:
Fixes dotnet/docs#5158

### DIFF
--- a/snippets/csharp/VS_Snippets_CLR_System/system.collections.generic.equalitycomparer/cs/program.cs
+++ b/snippets/csharp/VS_Snippets_CLR_System/system.collections.generic.equalitycomparer/cs/program.cs
@@ -4,105 +4,105 @@ using System.Collections.Generic;
 
 class Program
 {
-	static Dictionary<Box, String> boxes;
+    static Dictionary<Box, String> boxes;
 
-	static void Main()
-	{
-		BoxSameDimensions boxDim = new BoxSameDimensions();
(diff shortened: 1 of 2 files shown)


snippets/csharp/VS_Snippets_CLR_System/system.collections.generic.equalitycomparer/cs/program.cs
```diff
@@ -69,47 +69,33 @@ class BoxSameDimensions : EqualityComparer<Box>
...
-		}
-		else
-		{
+		else if (b1 == null || b2 == null)
 			return false;
-		}
+
+		return (b1.Height == b2.Height &&
+				b1.Length == b2.Length &&
+				b1.Width == b2.Width);
```
mikkelbu: I'm a bit unsure about the indentation here (and similar below for `BoxSameVolume`). It looked correct on my machine, but I also use 4 spaces for tabs (and the file is indented solely using tabs). I could remove one tab from each of the last 2 lines to make the comparison line up, but that will only work in places where tabs = 8 spaces.

mikkelbu:
@rpetrusha I've replaced the tabs with (4) spaces throughout the file.

Perhaps you should consider having a .editorconfig file in the root of the repository to help people following the code standard. This will of cause initially give you some larger diffs (due to tabs vs spaces etc), but it will make the samples more consistent on the long run.

PR #44, opened 2018-05-04
Pattern matching: aligned topic text with code

pkulikov:
Fixes dotnet/docs#5171

Also fixes triangle area.


svick:
This won't compile:

> GeometricUtilities.cs(34,25): error CS0077: The as operator must be used with a reference type or nullable type ('Square' is a non-nullable value type)
> GeometricUtilities.cs(38,25): error CS0077: The as operator must be used with a reference type or nullable type ('Circle' is a non-nullable value type)

What you need to do to fix that is to switch from `as` to a cast in the following code:

https://github.com/dotnet/samples/blob/6d0ec951f98f1890df7104ef4b6568147ce72b12/csharp/PatternMatching/GeometricUtilities.cs#L29-L46

pkulikov:
@svick thank you very much! I've fixed the code.

PR #46, opened 2018-05-05
List formatting in README

svick:


README.md
```diff
@@ -79,20 +79,20 @@ To create a sample:
             WhereClause1.MethodSyntaxExample();
         }
     }
-  ```
-  To build and run your sample...
+    ```
```
rpetrusha: Although "To build..." is now outdented so that it's not text attached to a bulleted list, the numbered list that follows starts with 5. I think that line 86 needs to be renumbered to 1.

svick: I did that on purpose. They are two related lists: you do what the first list says, then do what the second list says. Because of that, I think maintaining numbering between them makes sense.

Though I understand it's not a very common approach. Do you still think I should restart the numbering of the second list from 1?

rpetrusha: That would be my perference. Perhaps it's just me, but the number continuation makes the overall structure of the topic confusing. (It sort of plays with my mind.)

svick: @rpetrusha Ok, done.

PR #56, opened 2018-05-08
Added examples for What's new in .NET Core 2.1

rpetrusha:
## Added examples for What's new in .NET Core 2.1

//cc @mairaw

### DIFF
--- a/core/whats-new/whats-new-in-21/cs/brotli.cs
+++ b/core/whats-new/whats-new-in-21/cs/brotli.cs
@@ -12,6 +12,7 @@ public static class BrotliExample
         {
             decompressionStream.CopyTo(decompressedStream);
         }
+        decompressedStream.Position = 0;
         return decompressedStream;
     }
     // </Snippet1>
--- a/core/whats-new/whats-new-in-21/cs/brotli.cs
+++ b/core/whats-new/whats-new-in-21/cs/brotli.cs
@@ -1,0 +1,18 @@
+using System;
+using System.IO;
+using System.IO.Compression;
+
+public static class BrotliExample
+{
+    // <Snippet1>
+    public static Stream DecompressWithBrotli(Stream toDecompress)   
+    {
+        MemoryStream decompressedStream = new MemoryStream();
+        using (BrotliStream decompressionStream = new BrotliStream(toDecompress, CompressionMode.Decompress))
+        {
+            decompressionStream.CopyTo(decompressedStream);
+        }
+        return decompressedStream;
+    }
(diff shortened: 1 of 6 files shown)


core/whats-new/whats-new-in-21/cs/brotli.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    // <Snippet1>
+    public static Stream DecompressWithBrotli(Stream toDecompress)   
+    {
+        MemoryStream decompressedStream = new MemoryStream();
+        using (BrotliStream decompressionStream = new BrotliStream(toDecompress, CompressionMode.Decompress))
+        {
+            decompressionStream.CopyTo(decompressedStream);
+        }
+        return decompressedStream;
```
svick: Shouldn't you reset the `Position` of the `decompressedStream` to zero just before the `return`, so that caller can directly read the returned `Stream`?

core/whats-new/whats-new-in-21/vb/Program.vb
```diff
@@ -0,0 +1,7 @@
+Imports System
+
+Module Program
+    Sub Main(args As String())
+        Console.WriteLine("Hello World!")
+    End Sub
+End Module
```
svick: Do I understand it correctly that [VB doesn't support `Span<T>`](https://sharplab.io/#v2:DYLgbgRgNALuCmAnAlgMwJ4B8CSBbADgPaIwDOABAMrqkzy4CwAUAAoCuEwyAxuQLKEAJm2DxyLRIQDmiAIaMm5JeI5delDv1nIAdgAoAlM2UnyAEWS5yOtrghIKAXnIA5eAHdy2HXSlI9AKwADAbkAN4AvsamStExFlb4siTkzm6elEn6APKoXj7wfogGejZ2DlDkAIxBlTVGisoAojqCVBzMLW0CwqJAA=)? Though showing Hello World here doesn't seem like the right solution to me.

rpetrusha: The VB code won't show in the doc, @svick, except for the Brotli example. The VB code doesn't compile because the compiler sees Span(Of T) as a nested obsolete class.

svick: Then is there a reason to have the Program.vb here at all?

rpetrusha: Actually, no. I was leaving it as a placeholder for when Span<T> is supported in VB. But it's just as well to remove it.

PR #62, opened 2018-05-10
Csharp73 pattern based fixed statement

BillWagner:
Add a sample that uses the pattern based fixed statement with `Span<T>`

This makes use of the pattern `DangerousGetPinnableReference`

For dotnet/docs#4774

### DIFF
--- a/snippets/csharp/keywords/FixedKeywordExamples.cs
+++ b/snippets/csharp/keywords/FixedKeywordExamples.cs
@@ -52,9 +52,9 @@ namespace keywords
                 1,  5,  10, 10, 5,  1
             };
 
-            Span<int> RowSix = new Span<int>(PascalsTriangle, 10, 5);
+            Span<int> RowFive = new Span<int>(PascalsTriangle, 10, 5);
 
-            fixed (int* ptrToRow = RowSix)
+            fixed (int* ptrToRow = RowFive)
             {
                 // Use the numbers 1, 5, 10, 10, 5, 1
             }
--- a/snippets/csharp/keywords/FixedKeywordExamples.cs
+++ b/snippets/csharp/keywords/FixedKeywordExamples.cs
@@ -9,6 +9,7 @@ namespace keywords
         public static void Examples()
         {
             ModifyFixedStorage();
+            FixedSpanExample();
             InitializeFixedStorage();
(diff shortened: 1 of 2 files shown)


snippets/csharp/keywords/FixedKeywordExamples.cs
```diff
@@ -39,6 +40,27 @@ unsafe private static void ModifyFixedStorage()
...
+            int[] PascalsTriangle = {
+                          1,
+                        1,  1,
+                      1,  2,  1,
+                    1,  3,  3,  1,
+                  1,  4,  6,  4,  1,
+                1,  5,  10, 10, 5,  1
+            };
+
+            Span<int> RowSix = new Span<int>(PascalsTriangle, 10, 5);
```
pkulikov: It looks to be `RowFive` that starts at element 10 and contains 5 elements.
Or let's update argument values:
```csharp
Span<int> RowSix = new Span<int>(PascalsTriangle, 15, 6);
```

BillWagner: Yup, I changed the variable name.

snippets/csharp/keywords/FixedKeywordExamples.cs
```diff
@@ -39,6 +40,27 @@ unsafe private static void ModifyFixedStorage()
...
+                    1,  3,  3,  1,
+                  1,  4,  6,  4,  1,
+                1,  5,  10, 10, 5,  1
+            };
+
+            Span<int> RowFive = new Span<int>(PascalsTriangle, 10, 5);
+
+            fixed (int* ptrToRow = RowFive)
+            {
+                // Use the numbers 1, 5, 10, 10, 5, 1
```
pkulikov: Then here we can use numbers 1, 4, 6, 4, 1?

rpetrusha: The comment is wrong -- you're taking elements 10 through 14, which are 1,4,6,4,1.
Did you want to do anything with the numbers in the fixed block?

rpetrusha:
You can ignore my comments, @BillWagner -- you addressed them in your commit at about the same time I was submitting my comments. LGTM now.

PR #68, opened 2018-05-15
Csharp73 tuple equality

BillWagner:
Update the Tuples sample for the C# 7.3 tuple equality feature.

Samples for dotnet/docs#3965

### DIFF
--- a/snippets/csharp/tuples/tuples/Person.cs
+++ b/snippets/csharp/tuples/tuples/Person.cs
@@ -51,6 +51,41 @@ namespace tuples
 
 namespace TupleEquality
 {
+    public static class Tests
+    {
+        public static void EqualityTests()
+        {
+            Person nil = default;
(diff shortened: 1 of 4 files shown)


snippets/csharp/tuples/tuples/Person.cs
```diff
@@ -48,3 +48,76 @@ public static void Deconstruct(this Student s, out string first, out string last
...
+    public static class Tests
+    {
+        public static void EqualityTests()
+        {
+            Person nil = default;
+            Person blank = new Person(null, null);
+            Person sample = new Person("Bill", "Wagner");
+
+            Console.WriteLine("Starting Equality Tests");
+            Console.WriteLine($"nill equals nil: {nil == nil}");
```
rpetrusha: In this and the next two lines, `nill` should be `nil`

PR #74, opened 2018-05-19
Remove incorrect comment from Dispose pattern

svick:
In the Dispose pattern, if the base class doesn't have a finalizer, it shouldn't have any unmanaged objects, so it shouldn't have a comment showing where the non-existent unmanaged objects would be freed.


sharwell:
👍

svick:
@BillWagner 

Then you might be interested in https://github.com/dotnet/docs/issues/5485, which proposes a more radical change.

PR #78, opened 2018-05-24
Added snippets for the foreach keyword topic

pkulikov:
Supports dotnet/docs#5597


snippets/csharp/keywords/GenericWhereConstraints.cs
```diff
@@ -144,12 +144,15 @@ class Test<T, U>
     { }
     // </Snippet12>
 
-    // <Snippet13>
-    public class List<T>
+    namespace ListExample
```
pkulikov: Added that namespace to be able to use the generic .NET `List<T>` in the foreach snippets.

A separate namespace **per file** might be a better fix.

PR #80, opened 2018-05-25
update delegate and events samples

BillWagner:
For dotnet/docs#87

There were two major changes to update these samples.
1. UPgrade from project.json to csproj files.
1. Add snippet tags for including into the main topics.


csharp/delegates-and-events/FileLogger.cs
```diff
@@ -3,34 +3,38 @@
...
-                log.Flush();
+                using (var log = File.AppendText(logPath))
+                {
+                    log.WriteLine(msg);
+                    log.Flush();
+                }
+            }
+            catch (Exception)
+            {
+                // Hmm. We caught an exception while
```
rpetrusha: nit: since comments aren't localized, this might be better if it were in the example explanation rather than an inline comment.

BillWagner: There is explanatory text in the article as well. I'll leave the comment for those that just download the code.

PR #85, opened 2018-05-25
Added Span<T>.Enumerator examples

rpetrusha:
## Added Span<T>.Enumerator examples

//cc @BillWagner, @mairaw

### DIFF
--- a/snippets/csharp/api/system/span.enumerator/nosync/Program.cs
+++ b/snippets/csharp/api/system/span.enumerator/nosync/Program.cs
@@ -1,0 +1,46 @@
+﻿using System;
+using System.Threading;
+
+class Program
+{
+    private static readonly byte[] _array = new byte[5];
+
+    static void Main()
+    {
+        new Random(42).NextBytes(_array);
+        Span<byte> span = _array;
+
+        Thread thread = new Thread(delegate ()
+        {
+           ClearContents();
+        });
+
+        thread.Start();
+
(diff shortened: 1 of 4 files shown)


snippets/csharp/api/system/span.enumerator/nosync/Program.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        new Random(42).NextBytes(_array);
+        Span<byte> span = _array;
+
+        Thread thread = new Thread(delegate ()
+        {
+           ClearContents();
+        });
+
+        thread.Start();
```
pkulikov: Is using `Thread` API instead of `Task` intentional? Can we replace lines 13...18 with:
```csharp
Task.Run(ClearContents);
```

I understand we cannot make `async Main` and we cannot launch the span enumeration task. Also we might not want to use `Task.Delay(20).Wait()` instead of `Thread.Sleep(20)`. Thus, using `Thread` API for consistency everywhere.

BillWagner: I think we could use `async Main`, that's a released feature.

BillWagner: Good point. That's what I get for reviewing before :coffee:

BillWagner:
Looks good @rpetrusha You can :shipit: when ready.

PR #89, opened 2018-05-29
Update source.vb

the-wazz:
Moved the `<Snippet1>` marker to include the `Main` method. Without it, `TestCreateDataReader(DataTable dt)` is never called and the sample is incomplete, or doesn't do anything. (I'm assuming that `<Snippet1>` marks the start of what appears on the page.)

Another option would be to call `TestCreateDataReader(DataTable dt)` at the end of `GetCustomers()`, passing in the new table but that's not as clean.


BillWagner:
closing and re-opening to trigger CLA check.

the-wazz:
No problem. Take care.

PR #111, opened 2018-06-07
Fixed example of IEnumerable.Union<T>() method

M-Yankov:
## Summary  
Fixing the Snipped 4 (CSLINQEncapsulatedComparer#4) adding the comparer class to show correct results

## Documentation Link  
[docs.microsoft.com IEnumerable.Union](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.union)

Focus on the last snipped of first overload, the comparer is missing
```c#
    IEnumerable<ProductA> union =
         store1.Union(store2);
```

### DIFF
--- a/snippets/csharp/VS_Snippets_VBCSharp/CsLINQEncapsulatedComparer/CS/EncapsulatedComparer.cs
+++ b/snippets/csharp/VS_Snippets_VBCSharp/CsLINQEncapsulatedComparer/CS/EncapsulatedComparer.cs
@@ -89,7 +89,7 @@ class Program
         //excluding duplicates.
 
         IEnumerable<ProductA> union =
-          store1.Union(store2);
+          store1.Union(store2, new ProductComparer());
 
         foreach (var product in union)
             Console.WriteLine(product.Name + " " + product.Code);


BillWagner:
Thanks for contributing to docs @M-Yankov We appreciate it.

I've reviewed the changes, and I'm going to close this PR.  The sample you changed is for this [this overload](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.union?view=netframework-4.7.2#System_Linq_Enumerable_Union__1_System_Collections_Generic_IEnumerable___0__System_Collections_Generic_IEnumerable___0__) which should not include the comparer. The [second overload](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.union?view=netframework-4.7.2#System_Linq_Enumerable_Union__1_System_Collections_Generic_IEnumerable___0__System_Collections_Generic_IEnumerable___0__System_Collections_Generic_IEqualityComparer___0__) does include the comparer, and the sample shows that overload correctly.

M-Yankov:
Ok, but can you please look at the second example of [first overload](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.union?view=netframework-4.7.2#System_Linq_Enumerable_Union__1_System_Collections_Generic_IEnumerable___0__System_Collections_Generic_IEnumerable___0__)

Here is a [demo link dotnetfiddle](https://dotnetfiddle.net/VE16jI) 

In the documentation the output is:

```c#
/*
    This code produces the following output:
 
    apple 9
    orange 4
    lemon 12
*/
```

The demo output is:    
```
apple 9
orange 4
apple 9
lemon 12
```

Can you help explaining this discrepancies ?

BillWagner:
I'm going to need to dig into this a bit deeper, because there are a few interesting problems.

You're right that the code doesn't match the description. The included snippet file includes members for `Equals` and `GetHashCode`. But, those aren't included in the code and that means the 2nd overload is needed.

The other concern is that the example for the first overload shows the second overload. There's no example for the first overload.

Clearly something needs to change,. Let me dig a little deeper and I'll reply with a plan for what the right change is.

M-Yankov:
There is an example (first) shows using the first overload without comparer, which is OK:
```c#
int[] ints1 = { 5, 3, 9, 7, 5, 9, 3, 7 };
int[] ints2 = { 8, 3, 6, 4, 4, 9, 1, 0 };

IEnumerable<int> union = ints1.Union(ints2); 

foreach (int num in union)
{
    Console.Write("{0} ", num);
}

/*
 This code produces the following output:
 5 3 9 7 8 6 4 1 0
*/
```

BillWagner:
My apologies, @M-Yankov  You are correct.

I'll reopen and merge this.

M-Yankov:
Thanks! Glad to help.

PR #113, opened 2018-06-08
Add clarity to the pattern matching section of the F# Tour

baronfel:
On @csharpfritz's stream today there was some confusion around `id` in this section because there wasn't a clear link between a Person record and `id` in any way. I added some docs and renamed the parameter to hopefully highlight the links.


rpetrusha:
//cc @cartermp

PR #115, opened 2018-06-08
Added comment about control type

rpetrusha:
## Added comment about control type   

Fixes dotnet/docs#5863


rpetrusha:
Sorry, @mairaw. I thought that I'd fixed the misalignment, but clearly I hadn't. It's now fixed.

PR #124, opened 2018-06-15
Closing parenthesis placed incorrectly.

edwild-msft:
## Summary

SampleHost.exe was crashing on line 63 with a stack overflow in debug x64 build because i was getting assigned to a large number.  This change seems to let SampleHost.exe run correctly.


edwild-msft:
Thanks for looking at this.  The sample project is very helpful.

PR #143, opened 2018-06-27
Variable name MyStringBuilder -> myStringbuilder

majikandy:
## Summary

to make it clearer that myStringBuilder is the variable.

See also the docs that reference this, paragraph there also updated for the 1 mention of the instance variable name.

https://github.com/dotnet/docs/pull/6162


majikandy:
Yes, absolutely, I'll do that

PR #154, opened 2018-06-30
Use string interpolation in linq snippets

pkulikov:
Checked all the files in the [folder with C# LINQ snippets](https://github.com/dotnet/samples/tree/master/snippets/csharp/concepts/linq) to use string interpolation and fix its usage where necessary.

### DIFF
--- a/snippets/csharp/concepts/linq/how-to-order-the-results-of-a-join-clause_1.cs
+++ b/snippets/csharp/concepts/linq/how-to-order-the-results-of-a-join-clause_1.cs
@@ -66,7 +66,7 @@
                         Console.WriteLine(productGroup.Category);
                         foreach (var prodItem in productGroup.Products)
                         {
-                            Console.WriteLine("  {prodItem.Name:-10} {prodItem.CategoryID}");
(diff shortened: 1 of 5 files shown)


snippets/csharp/concepts/linq/how-to-order-the-results-of-a-join-clause_1.cs
```diff
@@ -66,7 +66,7 @@ select prod2
                         Console.WriteLine(productGroup.Category);
                         foreach (var prodItem in productGroup.Products)
                         {
-                            Console.WriteLine("  {prodItem.Name:-10} {prodItem.CategoryID}");
+                            Console.WriteLine($"  {prodItem.Name:-10} {prodItem.CategoryID}");
```
rpetrusha: this should be `{prodItem.Name,-10}`.

pkulikov: @rpetrusha that was a catch of the week! Thank you, fixed.

PR #160, opened 2018-07-05
async void should be async Task

BillWagner:
An async Main method must return Task or Task<int>

For dotnet/docs#5786


BillWagner:
Thanks @rpetrusha I'll add the second sample as well.

PR #163, opened 2018-07-05
Several uneccessary catch(Exception) were removed in c# examples.

jinek:
## Summary

Several catches were removed from c# examples (~15%)
No formatting for easier review. (Formatting will be added next PR)
VB and c++ examples will be fixed next.

Related to dotnet/docs#4870

### DIFF
--- a/snippets/csharp/VS_Snippets_ADO.NET/DataWorks DataTableLoad.IDataReader/CS/source.cs
+++ b/snippets/csharp/VS_Snippets_ADO.NET/DataWorks DataTableLoad.IDataReader/CS/source.cs
@@ -32,14 +32,8 @@ class Program
         // by the particular incompatibility:
         table = GetIntegerTable();
         reader = new DataTableReader(GetStringTable());
-        try 
-        {
-            table.Load(reader);
-        } 
-        catch (Exception ex) 
-        { 
-            Console.WriteLine(ex.GetType().Name + ":" + ex.Message);
-        }
+        
+        table.Load(reader);
 
         Console.WriteLine(" ============================= ");
         Console.WriteLine(
--- a/snippets/csharp/VS_Snippets_ADO.NET/DataWorks ConnectionStrings.Encrypt/CS/source.cs
+++ b/snippets/csharp/VS_Snippets_ADO.NET/DataWorks ConnectionStrings.Encrypt/CS/source.cs
@@ -12,8 +12,6 @@ class Program
     {
         // Takes the executable file name without the
         // .config extension.
-        try
-        {
             // Open the configuration file and retrieve 
             // the connectionStrings section.
             Configuration config = ConfigurationManager.
@@ -39,11 +37,6 @@ class Program
 
             Console.WriteLine("Protected={0}",
                 section.SectionInformation.IsProtected);
-        }
-        catch (Exception ex)
-        {
-            Console.WriteLine(ex.Message);
-        }
     }
     // </Snippet1>
 }
--- a/snippets/csharp/VS_Snippets_ADO.NET/DataWorks DataTableLoad.LoadOption/CS/source.cs
+++ b/snippets/csharp/VS_Snippets_ADO.NET/DataWorks DataTableLoad.LoadOption/CS/source.cs
@@ -33,15 +33,9 @@ class Program
         // by the particular incompatibility:
         table = GetIntegerTable();
         reader = new DataTableReader(GetStringTable());
-        try
-        {
-            table.Load(reader);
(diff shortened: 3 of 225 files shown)


Rick-Anderson:
Most of the ADO API's require try/catch for robust code. It's left to the app developer to handle the catch appropriately. Removing try/catch sends the wrong message.

snippets/csharp/VS_Snippets_ADO.NET/DataWorks DataTableLoad.IDataReader/CS/source.cs
```diff
@@ -32,14 +32,8 @@ static void Main()
         // by the particular incompatibility:
         table = GetIntegerTable();
         reader = new DataTableReader(GetStringTable());
-        try
```
Rick-Anderson: This should be caught - with the right DB exception.

jinek: If this message is for end-user, what is the idea we show "Out of memory" or "Object Null Reference"? If this is for developer, exception message will go to console output automatically. Removing catch also makes attached debugger stopped on the method that throws exception (for developer it should be more practical than to see only message in console?)

jinek: If this message is for end-user, what is the idea we show "Out of memory" or "Object Null Reference"? If this is for developer, exception message will go to console output automatically. Removing catch also makes attached debugger stopped on the method that throws exception (for developer it should be more practical than to see only message in console?)

jinek:
If this message is for end-user, what is the idea we show "Out of memory" or "Object Null Reference"? If this is for developer, exception message will go to console output automatically. Removing catch also makes attached debugger stopped on the method that throws exception (for developer it should be more practical than to see only message in console?)

jinek:
No problem about PR. But I think we are making a big mistake, the result of which looks something like this (very likely written by developer who is copypasting code from msdn):
image

I agree that developers copypaste the code. That is exactly why I supposed we should not show most improper way to deal with exceptions. Putting general catch does not make application very likely not to fail, it fails anyway, but additionally it gives a lot of big disadvantages: debugger won't stop, or won't be suggested to be attached. If it is an end-user application this exception won't be logged to system log and report won't be sent to developer. "Application closing with exitcode>0" is the default behavior that we miss here as well: caller of our application is thinking our application was executed successfully, but it was not and we left very less information answering why. Furthermore, in common case "catch" without of rethrowing opens the vulnerability - we are leaving code to execute in very unknown state. Of course in real world application we want to log information about exception, show a message to end-user and so on, but not only for exceptions from method Main: for global exception handling there are **AppDomain.UnhandledException** and **TaskScheduler.UnobservedTaskException** events. If our application is not being run behind balancer or any other kind of manager (like Windows Services) that can re-run our application into well-known state, and if we are sure it won't break related resources, we can do some tricks using **AppDomain.UnhandledException** to keep application running even in broken state, but probably this subject is out of database querying demonstration. 

Windows and .NET together already have default user-friendly and safe way to handle "unknown state", as I said, a system message "application stopped working" will be shown to user, information about exception will be automatically printed to console and standard "output", also it will be saved to windows events log, and, finally, application will be closed by system with proper exit code. Isn't it really enough for demonstration of code which is not related to exception handling?

I started with this only simple cases, I understand we need to show how to handle various exceptions that we can expect in different situations, but it will take a lot of time. Meanwhile may be instead of demonstrating unsafe/senseless way it is better not to show any way at all and just leave default behavior? We must not understand the result we don't expect, for example, OutOfMemoryException during calling "new DataTableReader()" or same, we must not understand "x" devided by zero, but currently we do. We demonstrating that developer should show any unknown message to user, but what should user do seeing "Division by zero" message?

Currently the very common situation can be found in almost any project (as what I saw 90%) looks similar to this:

```
try
{
   return _currentUser.Accounts.ById(accountId).Amount;
}
catch
{
  return 0;
}
```

that was probably copypasted from msdn as from a primary source of knowledges for learning developers.

I believe code we are discussing is the source of the problem I stated.

jinek:
@rpetrusha Probably you know about this, but there msdn has an article saying "don't do" like we are doing:  https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/exceptions/index 

> Do not catch an exception unless you can handle it and leave the application in a known state. If you catch System.Exception, rethrow it using the throw keyword at the end of the catch block. "

PR #165, opened 2018-07-08
Use classes instead of structs in pattern matching sample

svick:
I don't see much point in using structs here:

* It does not avoid any allocations, since values of these types are boxed right after creation.
* Classes are more commonly used in C#, so I think this makes the code easier to understand.

I have also considered making the classes `sealed`, but decided against it. It would probably be a good practice here, but it's adding a concept that does not really add anything to examples that are this simple.

### DIFF
--- a/csharp/PatternMatching/Shapes.cs
+++ b/csharp/PatternMatching/Shapes.cs
@@ -7,7 +7,7 @@ using System.Threading.Tasks;
 namespace PatternMatching
 {
     #region 01_ShapeDefinitions
-    public struct Square
+    public class Square
     {
         public double Side { get; }
 
@@ -16,7 +16,7 @@ namespace PatternMatching
             Side = side;
         }
     }
-    public struct Circle
+    public class Circle
     {
         public double Radius { get; }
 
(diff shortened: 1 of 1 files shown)


pkulikov:
@svick the [related topic](https://docs.microsoft.com/en-us/dotnet/csharp/pattern-matching) contains the following paragraph (above the first code snippet) that explains why structs are used:

> In this topic, you'll build a method that computes the area of different geometric shapes. But, you'll do it without resorting to object oriented techniques and building a class hierarchy for the different shapes. You'll use pattern matching instead. To further emphasize that we're not using inheritance, you'll make each shape a `struct` instead of a class. Note that different `struct` types cannot specify a common user defined base type, so inheritance is not a possible design. As you go through this sample, contrast this code with how it would be structured as an object hierarchy. When the data you must query and manipulate is not a class hierarchy, pattern matching enables very elegant designs.

If you still think that this PR should be merged, please update the topic text as well.

svick:
@pkulikov Thanks, I should have actually read the article.

I don't find that justification convincing: you don't have to use inheritance with classes and you can use a limited form of inheritance in the form of interfaces with structs.

So I still think classes are better here and have opened a PR to update that article.

PR #186, opened 2018-07-16
Added sample code to read attributes for C# tour

sputier:
This PR adds a snippet to demonstrate how to read attributes with C# at [C# Attributes - A tour of the C# language](https://docs.microsoft.com/en-us/dotnet/csharp/tour-of-csharp/attributes)

Fixes dotnet/docs#5424

### DIFF
--- a/snippets/csharp/tour/attributes/Program.cs
+++ b/snippets/csharp/tour/attributes/Program.cs
@@ -31,6 +31,7 @@
     {
         public static void Main(string[] args)
         {
+            // <SnippetReadAttributes>
             Type widgetType = typeof(Widget);
 
             //Gets every HelpAttribute defined for the Widget type
@@ -54,6 +55,7 @@
             }
 
             Console.ReadLine();
+            // </SnippetReadAttributes>
         }
     }
 }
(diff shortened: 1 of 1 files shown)


snippets/csharp/tour/attributes/Program.cs
```diff
@@ -31,6 +31,29 @@ public class Program
     {
         public static void Main(string[] args)
         {
+            Type widgetType = typeof(Widget);
```
BillWagner: Let's add a marker here (and at the end) rather than using line numbers. Line numbers are very brittle in the OPS build.  For example:

```csharp
// <SnippetReadAttributes>
```

snippets/csharp/tour/attributes/Program.cs
```diff
@@ -31,6 +31,29 @@ public class Program
...
+            //Gets every HelpAttribute defined for the Widget.Display method
+            object[] displayMethodAttributes = displayMethod.GetCustomAttributes(typeof(HelpAttribute), false);
+
+            if (displayMethodAttributes.Length > 0)
+            {
+                HelpAttribute attr = (HelpAttribute)displayMethodAttributes[0];
+                Console.WriteLine($"Display method help URL : {attr.Url} - Related topic : {attr.Topic}");
+            }
+
+            Console.ReadLine();
```
BillWagner: And then, after this line, close the snippet:

```csharp
// </SnippetReadAttributes>
```

PR #189, opened 2018-07-17
 Replaced occurences of 'the lazy fox jumped' with 'the lazy fox jumps'

sputier:
Fixes dotnet/docs#6412


snippets/visualbasic/VS_Snippets_CLR/Generic.LinkedList/vb/source.vb
```diff
@@ -157,48 +157,48 @@ End Class
...
+'the fox jumps over the dog 
 'Test 1: Add 'today' to beginning of the list:
-'today the fox jumped over the dog 
+'today the fox jumps over the dog 
 'Test 2: Move first node to be last node:
-'the fox jumped over the dog today 
+'the fox jumps over the dog today 
 'Test 3: Change the last node to 'yesterday':
-'the fox jumped over the dog yesterday 
+'the fox jumps over the dog yesterday
```
BillWagner: This is the only change that we might want to update. "The fox jumps over the dog yesterday" just sounds like it should be past tense.  Maybe the simplest change would be to replace "yesterday" with "later" so the grammar is correct.

sputier:
You're right ! I just fixed it :-)

PR #201, opened 2018-07-24
Samples for Unit Testing Best Practices Documentation

jpreese:
## Summary
Supporting samples for https://github.com/dotnet/docs/pull/6565

Describe your changes here.

This proved more difficult than anticipated. I really like the document structure that I have now. Comparing and contrasting better and worse test cases. For each comparison, taking the exact same inputs and just refactoring them to make them better.

In this compiled sample approach, I don't think that's *quite* possible as you'd end up with a lot of redundant tests and/or test cases that just don't make a ton of sense just for the sake of having a snippet. But maybe that's not a problem in this scenario?

I also took the liberty to rethink the examples given this new context (sidestepping the whole Glossary example). I chose something that has a single public method, and that could highlight all of the points I was getting across (magic strings, throwing exceptions, etc). It's a pretty well known Kata so should gain some familiarity points.

Looking for critiques/criticisms to move forward. We can stick with the glossary approach, take a different angle, scrap the thing, the world is our burrito. Thanks, team!


BillWagner:
@jpreese Comments below

> I really like the document structure that I have now. Comparing and contrasting better and worse test cases. For each comparison, taking the exact same inputs and just refactoring them to make them better.

The organization of the article does a great job of explaining the concepts. That's the primary goal. The samples are a great resource, but secondary to the goal of a clear, concise, article. (At least for our purposes here). Suggestions on the next paragraph.

> In this compiled sample approach, I don't think that's quite possible as you'd end up with a lot of redundant tests and/or test cases that just don't make a ton of sense just for the sake of having a snippet. But maybe that's not a problem in this scenario?

There are two approaches we use. For this sample, I'd recommend a *before* folder and an *after* folder. Then, you can include both as snippets. Also, you can link to the directories on GitHub for readers that want to download both versions, and experiment on their own. (For the record, our other strategy is to use inline snippets for "before" and the sample represents the finished work.)

> I also took the liberty to rethink the examples given this new context (sidestepping the whole Glossary example). I chose something that has a single public method, and that could highlight all of the points I was getting across (magic strings, throwing exceptions, etc). It's a pretty well known Kata so should gain some familiarity points.

I read the sample code you've got so far, and I really like it.

jpreese:
@BillWagner alright I went back and changed some things around. It looks like I was able to get it so the inputs lined up and it should go over pretty well in the core document. Let me know if you would like any other changes.

jpreese:
I think merging would be a good idea. It'd give us the _whole_ picture for the document itself. I'd want to swap out my markdown code with these snippets. 

If I notice anything that doesn't quite line up, I presume it should be easy enough to iterate on top of the samples.

PR #205, opened 2018-07-26
Added snippets for the nullable types topics

pkulikov:
Supports dotnet/docs#6676


snippets/csharp/programming-guide/nullable-types/nullable-ex1.cs
```diff
@@ -2,7 +2,7 @@
 
 class NullableExample
 {
-  static void Main()
+  static void Run()
```
pkulikov: I've made that change to keep the project buildable (only one `Main`). This snippet is no longer used in the updated conceptual docs. I'll delete it in another PR when move another snippets for nullable types topic in this repo. For now, it's kept here, so that this PR can be safely merged in without breaking dotnet/docs master.

PR #208, opened 2018-07-27
Examples of parsing ISO8601 strings

rpetrusha:
## Examples of parsing ISO8601 strings

Related to dotnet/docs#6331


snippets/csharp/api/system/datetimeoffset/parseexact/parseexact-iso8601-2.cs
```diff
@@ -0,0 +1,94 @@
...
+      ParseWithISO8601(dateStrings, DateTimeStyles.AssumeUniversal);   }
+
+   private static void ParseWithISO8601(string[] dateStrings, DateTimeStyles styles)
+   {   
+      Console.WriteLine($"Parsing with {styles}:");
+      foreach (var dateString in dateStrings)
+      {
+         try {
+            var date = DateTimeOffset.ParseExact(dateString, "O", null, styles);
+            Console.WriteLine($"   {dateString} --> {date:yyyy-MM-dd HH:mm:ss.FF zzz}");
```
BillWagner: really minor nit: Do you want to use a field width specifier on `dateString` to get them all to line up: `{dataString,-36}`

I think a field with of 36 works for every snippet in these examples.

PR #209, opened 2018-07-28
Updated the nullable type snippets

pkulikov:
Updated snippets for the dotnet/docs#6676

### DIFF
--- a/snippets/csharp/programming-guide/nullable-types/NullableTypesUsage.cs
+++ b/snippets/csharp/programming-guide/nullable-types/NullableTypesUsage.cs
@@ -6,7 +6,8 @@ namespace nullable_types
     {
         internal static void Examples()
         {
-            GetValueOfNullableType();
+            ExamineValueOfNullableType();
+            UseNullCoalescingOperator();
             ComparisonOperators();
         }
 
@@ -26,7 +27,7 @@ namespace nullable_types
             // </Snippet1>
         }
 
-        private static void GetValueOfNullableType()
+        private static void ExamineValueOfNullableType()
         {
             // <Snippet2>
             int? x = 10;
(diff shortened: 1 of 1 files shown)


snippets/csharp/programming-guide/nullable-types/NullableTypesUsage.cs
```diff
@@ -47,22 +51,23 @@ private static void AccessValueOfNullableType()
...
+        private static void ExplicitCast()
+        {
             // <Snippet5>
-            int? m;
-            int m2 = 10;
-            m = m2;  // Implicit conversion.
+            int? n = null;
+
+            //int m1 = n;    // Doesn't compile.
+            int n2 = (int)n; // Compiles, but throws an exception if n is null.
```
BillWagner: Now that pattern matching is a thing, I'd really like to see a sample that does something like this:

```csharp
  public static void Main()
  {
    int? thing = 5;
    int? notAThing = default;

    TestValue(thing);
    TestValue(notAThing);
  }
  private static void TestValue(int? nullableThing)
  {
    if (nullableThing is int number)
    {
      Console.WriteLine(number);
    } else
    {
      Console.WriteLine("thing is default value");
    }

  }
```

pkulikov:
@BillWagner thank you for suggestion! I've added the code example with pattern matching.

snippets/csharp/programming-guide/nullable-types/NullableTypesUsage.cs
```diff
@@ -47,41 +52,57 @@ private static void AccessValueOfNullableType()
...
-            int n2 = (int)n;   // Compiles, but throws an exception if n is null.
-            int n3 = n.Value;  // Compiles, but throws an exception if n is null.
+            int? z = 42;
+            if (z is int zValue)
+            {
+                Console.WriteLine($"z is {zValue}");
+            }
+            else
+            {
+                Console.WriteLine("z is undefined");
```
BillWagner: This is a great example, but I wouldn't use the word "undefined". Instead, I'd say "z does not have a value" or "z is null"

pkulikov: @BillWagner I've updated phrasing, also in other examples.

BillWagner:
Thanks again @pkulikov I've just :shipit: this, and I'll look at dotnet/docs#6676 now.

PR #223, opened 2018-08-04
fix samples for regex

herohua:
## Summary

Line 17 should have a space following `gray wolf`.


herohua:
@rpetrusha 

Thanks for your suggestion! Adding a quote does make output clearer. 
I've made the change accordingly. Please review again. :)

PR #228, opened 2018-08-08
Corrected loop counter in C# and C++ code

rpetrusha:
## Corrected loop counter in C# and C++ code

The example also uses a variable instead of retrieving a property value for a second time.

Fixes dotnet/docs#6887


BillWagner:
>LGTM but I can hear @BillWagner's voice telling you to also change the sample to use interpolated strings 😄

🎆 🥂 :100:

PR #237, opened 2018-08-14
Added using statement to dispose SqlDataReader, fixes 5914.

douglaslMS:
## Summary

Added a using statement to dispose the SqlDataReader (and other objects) per customer suggestion and best practices.

Fixes dotnet/docs#5914


snippets/csharp/VS_Snippets_ADO.NET/Classic WebData SqlDataReader.Close Example/CS/source.cs
```diff
@@ -21,21 +21,24 @@ private static void ReadOrderData(string connectionString)
...
+                        Console.WriteLine(String.Format("{0}, {1}",
+                            reader[0], reader[1]));
+                    }
 
-            // Call Close when done reading.
-            reader.Close();
+                    // Call Close when done reading.
+                   reader.Close();
+                }
+        }
```
BillWagner: This bracket should be indented by 4 spaces to line up correctly.

PR #247, opened 2018-08-19
Update and rename verbatim4.cs to verbatim2.cs

johnnyagerard:
This is a change for the 'docs/docs/csharp/language-reference/tokens/verbatim.md' file.

The goal is to simplify and use only one example instead of two.

### DIFF
--- a/snippets/csharp/language-reference/keywords/verbatim2.cs
+++ b/snippets/csharp/language-reference/keywords/verbatim2.cs
@@ -23,7 +23,7 @@ public class InfoAttribute : Attribute
    }
 }
 
-[@Info("A simple executable.")]
+[Info("A simple executable.")] // Prepend '@' to 'Info' to avoid compiler error CS1614.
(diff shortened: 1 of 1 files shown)


snippets/csharp/language-reference/keywords/verbatim2.cs
```diff
@@ -23,12 +23,12 @@ public InfoAttribute(string info)
    }
 }
 
-[@Info("A simple executable.")]
+[Info("A simple executable.")] // Prepend '@' to 'Info' to avoid compiler error CS1614.
```
BillWagner: I'd add a bit more to this comment.  I suggest the following:

```csharp
[Info("A simple executable.")] // Generates compiler error CS1614. Ambiguous Info and InfoAttribute. 
// Prepend '@' to select 'Info'. Specify the full name 'InfoAttribute' to select it.
```

PR #248, opened 2018-08-20
Add is constant pattern snippet for null check.

WayneGoosen:
## Summary

Added null check with `is` statement snippet WIP.

Fixes dotnet/docs#6855

### DIFF
--- a/snippets/csharp/language-reference/keywords/is/is-const-pattern11.cs
+++ b/snippets/csharp/language-reference/keywords/is/is-const-pattern11.cs
@@ -1,0 +1,25 @@
+// <Snippet11>
+using System;
+
+class Program
+{
+    static void Main(string[] args)
+    {
+        object o = null;
+
+        if (o is null)
+            Console.WriteLine("'is' constant pattern 'null' check : True");
+        
+        if (object.ReferenceEquals(o, null))
+            Console.WriteLine("object.ReferenceEqual 'null' check : True");
+
+        if (o == null)
+            Console.WriteLine("Equality operator (==) 'null' check : True");
+    }
+
+    // The example displays the following output:
+    //  'is' constant pattern 'null' check : True
+    //  object.ReferenceEqual 'null' check : True
+    //  Equality operator (==) 'null' check : True
+}
+// </Snippet11>
--- a/snippets/csharp/language-reference/keywords/is/is-const-pattern11.cs
+++ b/snippets/csharp/language-reference/keywords/is/is-const-pattern11.cs
@@ -8,18 +8,28 @@ class Program
         object o = null;
 
         if (o is null)
-            Console.WriteLine("'is' constant pattern 'null' check : True");
+        {
+            Console.WriteLine("o does not have a value");
+        }
+        else
+        {
+            Console.WriteLine($"o is {o}");
+        }
         
-        if (object.ReferenceEquals(o, null))
-            Console.WriteLine("object.ReferenceEqual 'null' check : True");
+        int? x = 10;
 
-        if (o == null)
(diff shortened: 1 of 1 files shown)


snippets/csharp/language-reference/keywords/is/is-const-pattern11.cs
```diff
@@ -0,0 +1,25 @@
...
+        object o = null;
+
+        if (o is null)
+            Console.WriteLine("'is' constant pattern 'null' check : True");
+        
+        if (object.ReferenceEquals(o, null))
+            Console.WriteLine("object.ReferenceEqual 'null' check : True");
+
+        if (o == null)
+            Console.WriteLine("Equality operator (==) 'null' check : True");
```
BillWagner: I recommend a second example here, using nullable types. A great way to use this syntax is something like:

```csharp
int? val = someMethodCall();

if (!(val is null))
{
   // Safely use val here.
}
```

pkulikov: @BillWagner with negation, the `is null` pattern looks less readable than using `!=` operator.

Either
```csharp
int? val = SomeMethodCall();
if (val is int v)
{
    // Use v here.
}
```

or
```csharp
int? val = SomeMethodCall();
if (val != null)
{
    // Safely use val.Value here.
}
```

seems to be more readable with nullable types than
```csharp
int? val = SomeMethodCall();
if (!(val is null))
{
    // Safely use val.Value here.
}
```

BillWagner: @pkulikov Yeah, that was a bad example.  maybe something more like:

```csharp
int? customerID  = findCustomerByName(firstName, lastName);
if (customerID  is null)
{
   customerID = registerCustomer(firstName, lastName);
}
```

pkulikov: @BillWagner that's a much better example; thank you.
I would nit pick about details: "Please register" in the raised message and immediate registration afterwards doesn't play well together, but idea is clear.

WayneGoosen:
Once i added the nullable type sample, it became clear i could write a consistent sample between reference and nullable type. Please let me know if you feel the comparison of null checks was valuable and I will bring it back into the sample

snippets/csharp/language-reference/keywords/is/is-const-pattern11.cs
WayneGoosen: I have the following as the sample now:

```csharp
object o = null;

if (o is null)
{
    Console.WriteLine("o does not have a value");
}
else
{
    Console.WriteLine($"o is {o}");
}

int? x = 10;

if (x is null)
{
    Console.WriteLine("x does not have a value");
}
else
{
    Console.WriteLine($"x is {x.Value}");
}

//  o does not have a value
//  x is 10
```
Once i added the nullable type sample, it became clear i could write a consistent sample between reference and nullable type. Please let me know if you feel the comparison of null checks was valuable and I will bring it back into the sample

BillWagner:
@WayneGoosen Thanks for updating this sample.  I like what you added.

> It became clear i could write a consistent sample between reference and nullable type. Please let me know if you feel the comparison of null checks was valuable and I will bring it back into the sample.


That's an important concept for readers to understand. It would be great to add this to the sample.  Once you're ready, remove the "wip" tag from the PR title, and I'll give it one final review and merge.

BillWagner:
Thanks for making those additional updates @WayneGoosen 

This looks great, and I'll :shipit: now.  You should see the changes on the live site on our next regular publishing cycle, no later than Monday morning.

Thanks again for contributing to docs!

PR #303, opened 2018-09-12
Example for IO exceptions article

rpetrusha:
## Example for IO exceptions article

Related to dotnet/docs#7385

### DIFF
--- a/snippets/standard/io/io-exceptions/cs/io-exceptions.cs
+++ b/snippets/standard/io/io-exceptions/cs/io-exceptions.cs
@@ -18,6 +18,7 @@ class Program
    {
         if (path is null) {
             Console.WriteLine("You did not supply a file path.");
+            return;
         }
 
         try {
--- a/snippets/standard/io/io-exceptions/vb/io-exceptions.vb
+++ b/snippets/standard/io/io-exceptions/vb/io-exceptions.vb
@@ -13,6 +13,7 @@ Module Program
    Function OpenStream(path As String) As StreamWriter
         If path is Nothing Then
             Console.WriteLine("You did not supply a file path.")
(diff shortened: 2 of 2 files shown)


snippets/standard/io/io-exceptions/cs/io-exceptions.cs
```diff
@@ -0,0 +1,54 @@
...
+            return;
+        sw.WriteLine("This is the first line.");
+        sw.WriteLine("This is the second line.");
+        sw.Close();
+    }
+
+   static StreamWriter OpenStream(string path)
+   {
+        if (path is null) {
+            Console.WriteLine("You did not supply a file path.");
```
BillWagner: Do you want to return here? As written, it will throw an exception later...

snippets/standard/io/io-exceptions/vb/io-exceptions.vb
```diff
@@ -0,0 +1,41 @@
...
+        If sw Is Nothing Then Return
+
+        sw.WriteLine("This is the first line.")
+        sw.WriteLine("This is the second line.")
+        sw.Close()
+    End Sub
+
+   Function OpenStream(path As String) As StreamWriter
+        If path is Nothing Then
+            Console.WriteLine("You did not supply a file path.")
```
BillWagner: Same comment as the C# sample regarding a return here.

snippets/standard/io/io-exceptions/cs/io-exceptions.cs
rpetrusha: Yes, @BillWagner, I did want to return. Thanks for pointing that out.

PR #308, opened 2018-09-15
Modernized and format example

lindexi:
## Summary

The previous code uses the first character lowercase method and the method uses `xx.GetType() == typeof(foo)` that can be instead of `is` and I modernize the code.

### DIFF
--- a/snippets/csharp/VS_Snippets_Wpf/DrawingMiscSnippets_snip/CSharp/EnumerateDrawingsExample.xaml.cs
+++ b/snippets/csharp/VS_Snippets_Wpf/DrawingMiscSnippets_snip/CSharp/EnumerateDrawingsExample.xaml.cs
@@ -5,14 +5,10 @@ using System.Windows.Media;
 using System.Windows.Media.Animation;
 using System.Windows.Navigation;
 
-
 namespace SDKSample
 {
-
-
-    public partial class EnumerateDrawingsExample: Page
+    public partial class EnumerateDrawingsExample : Page
     {
-
         public EnumerateDrawingsExample()
         {
             InitializeComponent();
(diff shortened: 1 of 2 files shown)


snippets/csharp/VS_Snippets_Wpf/DrawingMiscSnippets_snip/CSharp/EnumerateDrawingsExample.xaml.cs
```diff
@@ -5,65 +5,59 @@
...
+        // Enumerate the drawings in the DrawingGroup.
+        public void EnumDrawingGroup(DrawingGroup drawingGroup)
+        {
+            DrawingCollection dc = drawingGroup.Children;
+
+            // Enumerate the drawings in the DrawingCollection.
+            foreach (Drawing drawing in dc)
+            {
+                // If the drawing is a DrawingGroup, call the function recursively.
+                if (drawing is DrawingGroup @group)
```
BillWagner: I'm curious why you use `@group` instead of `group`

snippets/csharp/VS_Snippets_Wpf/DrawingMiscSnippets_snip/CSharp/EnumerateDrawingsExample.xaml.cs
```diff
@@ -5,65 +5,59 @@
...
+                    // Perform action based on drawing type.
+                }
+                else if (drawing is GlyphRunDrawing)
+                {
+                    // Perform action based on drawing type.
+                }
+                else if (drawing is VideoDrawing)
+                {
+                    // Perform action based on drawing type.
+                }
```
BillWagner: There are just enough combinations here to consider a pattern matching switch instead of the `is` statement.

What do you think @lindexi ?

lindexi: @BillWagner I also think `switch` is better than `is` but the original code has not used the converted value. I afraid my changed is too large and I am worried about modifying the code structure that I only change the `GetType` to `is`.

snippets/csharp/VS_Snippets_Wpf/DrawingMiscSnippets_snip/CSharp/EnumerateDrawingsExample.xaml.cs
lindexi: @BillWagner Thx for your review

snippets/csharp/VS_Snippets_Wpf/DrawingMiscSnippets_snip/CSharp/EnumerateDrawingsExample.xaml.cs
lindexi: @BillWagner Do you think it can be pattern matching switch?

BillWagner: @lindexi Either one works well. So, let's keep it as is.

I'll :shipit: now. 

Thanks for contributing to docs. 🎆

PR #318, opened 2018-09-21
Added calls to Task.Wait

rpetrusha:
## Added calls to Task.Wait

Two examples weren't working because they were exiting before the continuation completed.

Fixes dotnet/docs#7253


rpetrusha:
VB doesn't support `Async Main`. But the VB and C# examples don't have to be exactly parallel, so I've changed the C# examples to use `async main`, @BillWagner.

PR #343, opened 2018-10-03
Added examples for Memory<T> guidance

rpetrusha:
## Added examples for Memory\<T> guidance

Related to dotnet/docs#4823

### DIFF
--- a/snippets/standard/buffers/memory-t/owner/owner.cs
+++ b/snippets/standard/buffers/memory-t/owner/owner.cs
@@ -24,7 +24,7 @@ class Example
             Console.WriteLine($"You entered a number less than {Int32.MinValue:N0} or greater than {Int32.MaxValue:N0}.");
         }
         finally {
-            owner.Dispose();
+            owner?.Dispose();
         }
     }
 
--- a/snippets/standard/buffers/memory-t/owner/owner.cs
+++ b/snippets/standard/buffers/memory-t/owner/owner.cs
@@ -1,0 +1,43 @@
+﻿using System;
+using System.Buffers;
+
+class Example
+{
+    static void Main()
+    {
+        IMemoryOwner<char> owner = MemoryPool<char>.Shared.Rent();
+
+        Console.Write("Enter a number: ");
+	    try {
+            var value = Int32.Parse(Console.ReadLine());
+
+            var memory = owner.Memory;
+
+            WriteInt32ToBuffer(value, memory);
+
+            DisplayBufferToConsole(owner.Memory.Slice(0, value.ToString().Length));
+        }
+        catch (FormatException) {
+            Console.WriteLine("You did not enter a valid number.");
+        }
+        catch (OverflowException) {
+            Console.WriteLine($"You entered a number less than {Int32.MinValue:N0} or greater than {Int32.MaxValue:N0}.");
+        }
+        finally {
+            owner.Dispose();
+        }
+    }
+
+    static void WriteInt32ToBuffer(int value, Memory<char> buffer)
+    {
+        var strValue = value.ToString();
+
+        var span = buffer.Span;
+        for (int ctr = 0; ctr < strValue.Length; ctr++)
+            span[ctr] = strValue[ctr];
+    }
+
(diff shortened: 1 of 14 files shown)


snippets/standard/buffers/memory-t/owner/owner.cs
```diff
@@ -0,0 +1,43 @@
...
+            DisplayBufferToConsole(owner.Memory.Slice(0, value.ToString().Length));
+        }
+        catch (FormatException) {
+            Console.WriteLine("You did not enter a valid number.");
+        }
+        catch (OverflowException) {
+            Console.WriteLine($"You entered a number less than {Int32.MinValue:N0} or greater than {Int32.MaxValue:N0}.");
+        }
+        finally {
+            owner.Dispose();
```
BillWagner: minor nit:  I'd prefer `owner?.Dispose();` to protect against the statement in line 8 throwing.

snippets/standard/buffers/memory-t/task-returning/task-returning.cs
```diff
@@ -0,0 +1,55 @@
...
+using System;
+using System.Buffers;
+using System.IO;
+using System.Threading.Tasks;
+
+public class Example
+{
+    // <Snippet1>
+    // An acceptable implementation.
+    static void Log(ReadOnlyMemory<char> message)
```
BillWagner: I'd prefer seeing this method return the `Task` so that it can be awaited.

Unless, what you're showing is how `ReadOnlyMemory` is limited in it use because it is a `ref struct`. I didn't compile this code, but I think this use is allowed, but the caller may be limited.

snippets/standard/buffers/memory-t/task-returning/task-returning.cs
```diff
@@ -0,0 +1,55 @@
...
+            var memory = owner.Memory;
+            var span = memory.Span;
+            while (true)
+            {
+                int value = Int32.Parse(Console.ReadLine());
+                if (value < 0) 
+                    return;
+
+                int numCharsWritten = ToBuffer(value, span);
+                Log(memory.Slice(0, numCharsWritten));
```
BillWagner: If `Log` returns a `Task`, this can be awaited (and main can be an async main method).

snippets/standard/buffers/memory-t/void-returning-async/void-returning-async.cs
```diff
@@ -0,0 +1,56 @@
...
+using System;
+using System.Buffers;
+using System.IO;
+using System.Threading.Tasks;
+
+public class Example
+{
+    // <Snippet1>
+    // An acceptable implementation.
+    static void Log(ReadOnlyMemory<char> message)
```
BillWagner: Same comment on async as the previous samples.

snippets/standard/buffers/memory-t/task-returning/task-returning.cs
rpetrusha: That won't work, since spans aren't allowed in async methods.

snippets/standard/buffers/memory-t/void-returning-async/void-returning-async.cs
rpetrusha: That won't work, since spans aren't allowed in async methods.

pkulikov: @rpetrusha @BillWagner does the code use [this `ReadOnlyMemory` struct](https://github.com/dotnet/corefx/blob/master/src/Common/src/CoreLib/System/ReadOnlyMemory.cs)? Then it's not `ref struct`, but usual `readonly struct`. Then, I expect it can be used in `async` methods. Or I'm wrong here?

pkulikov: Oh, I see: though the `Log` might be `async`, it, anyway, cannot be awaited in the `Main` as `Main` contains spans and cannot be `async`. And not awaiting `async` methods might be not a good example.

PR #351, opened 2018-10-06
add Memory<T> for VB.Net sample

binh0001:
add Memory<T> for VB.Net sample

### DIFF
--- a/core/whats-new/whats-new-in-21/vb/Program.vb
+++ b/core/whats-new/whats-new-in-21/vb/Program.vb
@@ -1,0 +1,17 @@
+﻿Module Program
+    Sub Main()
+        Dim numbers As Integer() = New Integer(99) {}
+
+        For i As Integer = 0 To 99
+            numbers(i) = i * 2
+        Next
+
+        Dim part = New Memory(Of Integer)(numbers, start:=10, length:=10)
+
+        For Each value In part.Span
+            Console.Write($"{value}  ")
+        Next
+    End Sub
+End Module
+' The example displays the following output:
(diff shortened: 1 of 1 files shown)


core/whats-new/whats-new-in-21/cs/Program.cs
```diff
@@ -13,6 +13,8 @@ static void Main()
         var part = new Span<int>(numbers, start: 10, length: 10);
         foreach (var value in part)
             Console.Write($"{value}  ");
+
+        Console.ReadLine();
```
rpetrusha: We generally don't include calls to Console.ReadLine; it's useful when running in debug mode in Visual Studio, but not when run outside of it. (And without a message, the need to press Enter to end the example is not obvious.) A call to Console.WriteLine makes sense, though, since the previous statement calls Console.Write.

core/whats-new/whats-new-in-21/vb/Program.vb
```diff
@@ -0,0 +1,19 @@
...
+            numbers(i) = i * 2
+        Next
+
+        Dim part = New Memory(Of Integer)(numbers, start:=10, length:=10)
+
+        For Each value In part.Span
+            Console.Write($"{value}  ")
+        Next
+
+        Console.ReadLine()
```
rpetrusha: See the remark about Console.ReadLine in the C# example.

rpetrusha:
Thanks for making the additional changes, @binh0001. I'll merge your PR now. Let us know whether you'll open the PR to include the example in the What's New article in dotnet/docs, or whether we should open the PR.

PR #355, opened 2018-10-06
Add multithreading to ConcurrentBag example

jeremyjung:
## Summary
Adds multithreading to the ConcurrentBag example.  Seeking feedback on if this is the type of example you are looking for and if you would also like multithreaded reads from the collection.

Partially addresses dotnet/docs#7979


rpetrusha:
Thanks, @jeremyjung, for making these changes, and especially for moving the examples. I've approved your PR, and I'll merge it now.

PR #361, opened 2018-10-09
Add "useHeader: true" to TextLoader.CreateFrom

zhiliaxu:
## Summary

Since the tsv files contain header line, we need to pass `useHeader: true` to `TextLoader.CreateFrom` method.


rpetrusha:
@JRAlexander, could you take a look, please?

JRAlexander:
Thanks, @zhiliaxu, for this contribution, I really appreciate your passion. Unfortunately, the sample needs to stay as is.
The reason we didn't use useHeader: true is because we were calling out the Column ordinal attribute in the tutorial specifically. In the SentimentData class, both fields have Column attributes attached to them. This attribute describes the order of each field in the data file, and which is the Label field.  
So really, the sample is fine, but the reasoning needs to be more explicit in the tutorial. We will revise it when we update for the API changes. Thanks again!

PR #362, opened 2018-10-09
WPF getting started instructions

diverdan92:
Adding WPF getting started instructions on top of .NET Core


rido-min:
I'm thinking of including different strategies to migrate the head project:

- Linked files from the original project
- Linked files to the original project
- Copy files from the original project
- Project reference to the original project

Adding some screenshots will help to understand those.

PR #377, opened 2018-10-13
Remove unecessary try..catch constructs

Mackiovello:
Fixes https://github.com/dotnet/docs/issues/6271


Mackiovello:
I decided to not do anything about the wonky spacing and indenting in many of these files. That should make it easier to review.

Mackiovello:
I decided to always remove the `try..catch` constructs since they always just swallowed the exception and wrote to the console. There are very few cases where that's a good practice.

BillWagner:
@Mackiovello I know this is aging a bit.  

I'm having some internal discussions about what's the best plan here, because our sample code is used as a learning tool. Sometimes that means we use different practices than we would recommend for a production code base. We'll comment / review formally soon.

snippets/csharp/VS_Snippets_CLR_System/system.Security.Permissions.UrlIdentityPermission/CS/urlidentity.cs
```diff
@@ -51,29 +51,17 @@ private static void UnionDemo()
         UrlIdentityPermission permIdPerm1 = new UrlIdentityPermission("http://www.fourthcoffee.com/process/");
         UrlIdentityPermission permIdPerm2 = new UrlIdentityPermission("http://www.fourthcoffee.com/*");
         UrlIdentityPermission p3 = (UrlIdentityPermission)permIdPerm1.Union(permIdPerm2);
```
rpetrusha: This is throwing a NotSupportedException. Despite the documentation, this seems to only succeed if given two equal URLs. At the same time, the functionality is not supported in .NET Framework 4 and later. But the example is not used. The UnionDemo method can be removed from the source code (and the call to UnionDemo on line 12 should also be removed).

snippets/csharp/VS_Snippets_Winforms/ContainerExmpl/CS/librarycontainer.cs
```diff
@@ -228,22 +228,15 @@ static void Main(string[] args)
...
-				cntrExmpl.Add(book3, "0812551478");
-				BookComponent book4 = new BookComponent("The Soul of the Fire", "Terry Gooodkind");
-				//This will generate exception because the ISBN already exists in the container.
-				cntrExmpl.Add(book4, "0812551478");
-			}
-			catch(SystemException e)
-			{
-				Console.WriteLine("Error description: " + e.Message);
-			}
+			BookComponent book1 = new BookComponent("Wizard's First Rule", "Terry Gooodkind");
```
rpetrusha: Since the container doesn't support duplicates, we can't simply not handle exceptions. The problem here is that line 184 throws a SystemException, which is inappropriate. You could change it to ArgumentException, then handle the ArgumentException in a try/catch block around all of the BookComponent instantiates and calls to the Add method.

Mackiovello:
Thank you so much for the detailed feedback @rpetrusha!

I'll go through the examples over the next couple of days and ping you when I'm done.

I'm not quite sure how the project files are structured. For some examples, there's a project file in the same directory, but for others, there doesn't seem to be one. For example, for the `GetProcessByName`, I had to add my own project file in order to build and run. How is it supposed to work?

Also, I'm not super familiar with VB and C++, so please be extra observant that I'm following best practices with those languages.

Mackiovello:
Checklist for myself:

__GetProcessByName__  
- [ ] VB
- [ ] C#
- [ ] C++

__UrlIdentity__
- [ ] VB
- [ ] C#
- [ ] C++

__LibraryContainer__
- [ ] VB
- [ ] C#
- [ ] C++

rpetrusha:
Thanks for responding, @Mackiovello. It's nice to be working with you again.
 
Since most of this is legacy code (almost all of the examples that handle Exception, and all of the example that handle SystemException, were written for .NET Framework 1.0 or 1.1), the examples were tested and run using the command-line compiler only. (In fact, when the examples were created, MSBuild didn't exist.) They can still be run using csc.exe.

Since our C++/CLI coverage tends to be spotty, we can simply plan on removing the C++ examples from the documentation if you don't feel comfortable modifying them. If you have difficulty with any of the VB examples, I can do them.

Mackiovello:
> Since our C++/CLI coverage tends to be spotty, we can simply plan on removing the C++ examples from the documentation if you don't feel comfortable modifying them.

I could modify the C++ examples, but I don't have an environment setup for it so it would take some extra time. Should I create an issue to remove them or remove them in this PR?

rpetrusha:
You can just remove them in this PR, @Mackiovello. We want to leave existing C++ examples that don't require modification.

rpetrusha:
Thank you so much, @Mackiovello, both for submitting the PR and for all of the additional changes you've made. This looks really good, so I'll merge your PR now.

PR #397, opened 2018-10-20
Fix style in Regex.Replace example

kayKayEhnn:
## Summary
Apply stylistic changes to coincide with other examples'.


rpetrusha:
Thank you, @kayKayEhnn, for adding the Visual Basic example and using interpolated strings in both examples. I did remove the `Imports System` statement from the Visual Basic example, since it is not needed. I'll merge your PR now. Again, thanks.

PR #403, opened 2018-10-21
Prefer async and await with Tasks rather than blocking

benaadams:
## Summary

Otherwise people copying the samples can cause other problems depending on where they are used, like threadpool starvation if the blocking happens on threadpool threads (which is likely with Tasks)


benaadams:
There are quite a few more to change; but will wait for feedback before continuing

BillWagner:
@benaadams Thanks for submitting all these.

I had intended to review them this morning, but the GitHub service outage was giving me pause. I will review them all tomorrow morning.

BillWagner:
closing and reopening to trigger a status check.

PR #411, opened 2018-10-24
Updating .NET Core 3 SDK download link

diverdan92:
Changing download link for netcore3 sdk to a vanity url so we can update it quickly to recommended SDKs as we continue making progress (i.e. eventually for preview releases)


diverdan92:
Thanks, @rpetrusha

PR #422, opened 2018-10-28
await function calls

imba-tjd:
I referred to https://github.com/dotnet/samples/pull/318, but I don't know vb. Please directly modify them.
Fixes https://github.com/dotnet/docs/issues/8652.

### DIFF
--- a/snippets/csharp/VS_Snippets_CLR_System/system.IO.StreamReader/CS/asyncex1.cs
+++ b/snippets/csharp/VS_Snippets_CLR_System/system.IO.StreamReader/CS/asyncex1.cs
@@ -5,12 +5,12 @@ using System.Threading.Tasks;
 
 class Example
 {
-    public static void Main()
+    static async Task Main()
     {
-        ReadAndDisplayFilesAsync();
+        await ReadAndDisplayFilesAsync();
     }
 
-    private async static void ReadAndDisplayFilesAsync()
+    static async Task ReadAndDisplayFilesAsync()
     {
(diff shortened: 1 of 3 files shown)


BillWagner:
@rpetrusha @KathleenDollard I think the right solution is to use `GetAwaiter().GetResult()` in the `Main` sub:

```vb
Imports System.IO
Imports System.Threading.Tasks

Module Module1

    Sub Main()
        ReadCharacters().GetAwaiter().GetResult()
    End Sub

    Async Function ReadCharacters() as Task
        Dim result As String

        Using reader As StreamReader = File.OpenText("existingfile.txt")
            Console.WriteLine("Opened file.")
            result = Await reader.ReadLineAsync()
            Console.WriteLine("First line contains: " + result)
        End Using
    End Function
End Module
```

Is there an easier VB idiom?

KathleenDollard:
I want to run that by the community. I'm fine to merge it and then see if they have a better approach, or to wait. That is up to you. 

[VBLang question on this](https://github.com/dotnet/vblang/issues/352)

rpetrusha:
@BillWagner, @KathleenDollard, instead of GetAwaiter.GetResult, isn't Wait simpler?

I think, though, that we should merge this PR and but address the VB code separately in another PR -- it generally doesn't produce any output, since Main completes before the async method has completed.

BillWagner:
I'll :shipit: with the C# samples.

See dotnet/VBLang#352 for thoughts on the VB samples.

PR #425, opened 2018-10-30
Remove condition unnecessary

lucasnobredev:
## Summary

This PR is from https://github.com/dotnet/docs/issues/8727.

Fixes dotnet/docs#8727


BillWagner:
@lucasnobredev I added the WIP label to this PR.

Rather than remove this code, I'd like us to come up with a better sample that shows the value of discards in pattern matching scenarios.  I don't have really good ones top of mind, but I'll chat a bit with @rpetrusha and others to make some suggestions.

PR #426, opened 2018-10-30
Fix strings with s uppercase

lucasnobredev:
## Summary

This PR is from https://github.com/dotnet/docs/issues/8700.

Fixes dotnet/docs#8700

### DIFF
--- a/snippets/csharp/programming-guide/discards/discard-pattern2.cs
+++ b/snippets/csharp/programming-guide/discards/discard-pattern2.cs
@@ -21,7 +21,7 @@ public class Example
          Console.Write("A null object reference: ");
          Console.WriteLine("Its use could result in a NullReferenceException");
       }
-      else if (obj is var _)
+      else
          Console.WriteLine($"Some object type without format information");
    }
 }
--- a/snippets/csharp/programming-guide/discards/discard-pattern2.cs
+++ b/snippets/csharp/programming-guide/discards/discard-pattern2.cs
@@ -21,7 +21,7 @@ public class Example
          Console.Write("A null object reference: ");
          Console.WriteLine("Its use could result in a NullReferenceException");
       }
-      else
+      else if (obj is var _)
          Console.WriteLine($"Some object type without format information");
    }
 }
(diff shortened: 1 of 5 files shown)


snippets/csharp/programming-guide/discards/discard-pattern2.cs
```diff
@@ -21,7 +21,7 @@ private static void ProvidesFormatInfo(object obj)
          Console.Write("A null object reference: ");
          Console.WriteLine("Its use could result in a NullReferenceException");
       }
-      else if (obj is var _)
+      else
```
BillWagner: Let's leave this sample as is for the moment.  Issue dotnet/docs#8727 points to the need for a  better sample. I'd rather leave this in place until we make a new sample that shows the usage better.

rpetrusha: Although this change is acceptable, it requires a change to example description. The example is used in [Discards - C# Guide](https://docs.microsoft.com/en-us/dotnet/csharp/discards), where the sentence "It also used the discard pattern to handle non-null objects of any other type." needs to be removed. Could you open a PR to do that?

rpetrusha:
@BillWagner and I reviewed this at about the same time, @lucasnobredev. I agree with him, though, that the final example should be left in place until we come up with a better example. Since you've also made the same update in #425, the last change can be reverted.

lucasnobredev:
@rpetrusha @BillWagner I was reading the article again and it seems like if we remove condition of code and remove the phrase "It also used the discard pattern to handle non-null objects of any other type", the section lost the sense to exist. I'll take this comment to https://github.com/dotnet/docs/issues/8727 and I'll wait.

BillWagner:
I looked at the latest updates, and I'm ready to :shipit:

Let me know what you think @rpetrusha

rpetrusha:
I agree that this is ready to merge, @BillWagner. Thanks, @lucasnobredev. I've tagged this as verify build before merge, since the merge to live build is still running. It can be merged once that build completes and the merge-to-live PR is closed.

PR #428, opened 2018-10-30
Use the appropriate loop and bracket on newline

Mackiovello:
The code didn't follow C# coding conventions by using bracket on the same line, also, `foreach` fit better in this case.

### DIFF
--- a/snippets/csharp/language-reference/keywords/await/await2.cs
+++ b/snippets/csharp/language-reference/keywords/await/await2.cs
@@ -20,11 +20,13 @@ class Example
       var client = new HttpClient();
       long pageLengths = 0;
 
-      for (int ctr = 1; ctr < args.Length; ctr++) {
-         var uri = new Uri(Uri.EscapeUriString(args[ctr]));
+      foreach (var arg in args)
+      {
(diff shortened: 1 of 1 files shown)


snippets/csharp/language-reference/keywords/await/await2.cs
```diff
@@ -20,11 +20,13 @@ private static async Task<long> GetPageLengthsAsync(string[] args)
       var client = new HttpClient();
       long pageLengths = 0;
 
-      for (int ctr = 1; ctr < args.Length; ctr++) {
-         var uri = new Uri(Uri.EscapeUriString(args[ctr]));
+      foreach (var arg in args)
+      {
+         var uri = new Uri(Uri.EscapeUriString(arg));
```
pkulikov: @Mackiovello note that the `for` loop variable is initialized at 1 (`ctr = 1`). That means that the first element of `args` is skipped.

Mackiovello: Dang, that's sloppy by me. Thanks for noticing. I'll fix it.

pkulikov:
@Mackiovello good change! What do you think about making the `Main` method async and use await call inside it, not the Result property?

PR #468, opened 2018-11-19
Updated the conditional operator examples

pkulikov:
Supports dotnet/docs#9125

### DIFF
--- a/snippets/csharp/language-reference/operators/ConditionalExamples.cs
+++ b/snippets/csharp/language-reference/operators/ConditionalExamples.cs
@@ -1,6 +1,4 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.Text;
 
 namespace operators
 {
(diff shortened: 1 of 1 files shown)


snippets/csharp/language-reference/operators/ConditionalExamples.cs
```diff
@@ -1,45 +1,67 @@
...
             var largeArray = new int[] { 10, 20, 30, 40, 50 };
 
-            int index= new Random().Next(0, 9);
+            int index = 7;
             ref int refValue = ref ((index < 5) ? ref smallArray[index] : ref largeArray[index - 5]);
-            Console.WriteLine(refValue);
+            refValue = 0;
+
+            index = 2;
+            ((index < 5) ? ref smallArray[index] : ref largeArray[index - 5]) = 100;
```
pkulikov: To show that `ref` in front of the conditional ref expression is not necessary.
Also, I've made the examples non-random, in order to present the output in comments. Examples with `ref` doesn't run in the interactive mode.

snippets/csharp/language-reference/operators/ConditionalExamples.cs
```diff
@@ -1,45 +1,67 @@
...
         }
 
-
-        static void ConditionalValueExpressions()
+        private static void ConditionalValueExpressions()
         {
             // <SnippetConditionalValue>
-            double sinc(double x) =>
-                x != 0.0 ? Math.Sin(x) / x : 1.0;
+            double sinc(double x) => x != 0.0 ? Math.Sin(x) / x : 1;
```
pkulikov: Minor tweak to show that types might be different (`double` and `int` in this case).

PR #490, opened 2018-12-02
Correctly dispose HttpClient in its example

svick:
Fixes https://github.com/dotnet/dotnet-api-docs/issues/1045.

Also switches from explicitly calling `Dispose` to `using`, because I think that's a better practice.

In the process of testing this change, I used the following csproj:

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFrameworks>net45;netcoreapp1.0</TargetFrameworks>
    <LangVersion>7.1</LangVersion>
  </PropertyGroup>

  <ItemGroup Condition="'$(TargetFramework)' == 'net45'">
    <Reference Include="System.Net.Http" />
  </ItemGroup>

</Project>
```

Would it be useful if I included it in this PR?

### DIFF
--- a/snippets/csharp/VS_Snippets_Misc/system.net.http.httpclient/cs/source.cs
+++ b/snippets/csharp/VS_Snippets_Misc/system.net.http.httpclient/cs/source.cs
@@ -9,29 +9,26 @@ class HttpClient_Example
 // <Snippet1>
    static async Task Main()
    {
-      // Create a New HttpClient object.
-      HttpClient client = new HttpClient();
-    
-      // Call asynchronous network methods in a try/catch block to handle exceptions
-      try	
+      // Create a New HttpClient object and dispose it when done, so the app doesn't leak resources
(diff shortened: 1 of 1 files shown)


svick:
@richlander 

1. Isn't that specific to ASP.NET Core? I mean, it's technically usable outside of it, but is using it actually a recommended practice in, say, WPF or Xamarin?
2. Wouldn't that require lots of dependency injection-related code? I think that doesn't belong here.

Though adding a link to a document about it might make sense.

BillWagner:
@svick: 

> Would it be useful if I included it (the csproj) in this PR?

What we've been doing as time permits is migrating the snippets into the dotnet/samples repository, and making them part of a buildable project. That does require 2 PRs: one in samples to create the new sample, and a separate one in dotnet/docs to refer to the new sample and remove the old code. I didn't want to hold up a good PR on that process though.

PR #493, opened 2018-12-03
Added snippets for true and false operators/literals

pkulikov:
Supports dotnet/docs#9356

### DIFF
--- a/snippets/csharp/keywords/TrueFalseOperatorsExample.cs
+++ b/snippets/csharp/keywords/TrueFalseOperatorsExample.cs
@@ -13,7 +13,7 @@ public struct LaunchStatus
         this.status = status;
     }
 
-    public static bool operator true(LaunchStatus x) => x == Green;
+    public static bool operator true(LaunchStatus x) => x == Green || x == Yellow;
     public static bool operator false(LaunchStatus x) => x == Red;
 
     public static LaunchStatus operator &(LaunchStatus x, LaunchStatus y)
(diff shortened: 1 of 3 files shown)


snippets/csharp/keywords/TrueFalseOperatorsExample.cs
```diff
@@ -0,0 +1,55 @@
...
+    public static bool operator false(LaunchStatus x) => x == Red;
+
+    public static LaunchStatus operator &(LaunchStatus x, LaunchStatus y)
+    {
+        if (x == Red || y == Red || (x == Yellow && y == Yellow))
+        {
+            return Red;
+        }
+
+        return Green;
```
BillWagner: I think there's an issue here. For `x & y & z` work correctly when `z` is `Yellow`, this should return `Yellow` when one of `x` or `y` is `Yellow`.  

Do you agree?

pkulikov: @BillWagner 
> For `x & y & z` work correctly...

to clarify: by "correctly" you mean that the `&` overload must be associative, so `(x & y) & z == x & (y & z)`

Yes, I agree then. Will make an update soon.

pkulikov: @BillWagner I've addressed your feedback; thanks for review.

I've checked the updated definition for all possible `x`, `y`, and `z` values: now the `&` operator is commutative and associative. In particular, that means that `x1 & ... & xN` is `Yellow` iff there is only one operand which is `Yellow` with all others being `Green`. That's why I've updated the `true` definition to return `true` value for `Yellow` operand. The following requirement is met:
> if only one sub-system is yellow, we can launch.

BillWagner:
Thanks for making these updates. I'll :shipit: now and then check the article PR.

PR #505, opened 2018-12-07
Added snippets for the division operator article

pkulikov:
Supports dotnet/docs#9486

### DIFF
--- a/snippets/csharp/language-reference/operators/DivisionExamples.cs
+++ b/snippets/csharp/language-reference/operators/DivisionExamples.cs
@@ -29,7 +29,7 @@ namespace operators
 
             int a = 13;
             int b = 5;
-            Console.WriteLine(a / (double)b);  // output: 2.6
+            Console.WriteLine((double)a / b);  // output: 2.6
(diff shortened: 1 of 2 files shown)


snippets/csharp/language-reference/operators/DivisionExamples.cs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        private static void IntegerAsFloatingPointDivision()
+        {
+            // <SnippetIntegerAsFloatingPoint>
+            Console.WriteLine(13 / 5.0);       // output: 2.6
+
+            int a = 13;
+            int b = 5;
+            Console.WriteLine(a / (double)b);  // output: 2.6
```
BillWagner: I'd consider casting `a` instead of `b` to show that if *either* the numerator or the denominator is a double, the result is a double. (The first sample has an integer numerator divided by a floating point denominator).

BillWagner:
Thanks @pkulikov 

I'll :shipit: now, and then restart a build on the docs PR and check that one.

PR #521, opened 2018-12-15
Source code for updated clustering tutorial

pkulikov:
The source code for the updated ML.NET clustering tutorial. Uses a new "dynamic" API.

Rather than edit existing sample code:
https://github.com/dotnet/samples/tree/master/machine-learning/tutorials/IrisClustering

I've decided to use it as the starting point for a new solution presented in this PR. 

Thus, this PR is safe to merge to master. It wouldn't break any existing references in the dotnet/docs repo.

Supports dotnet/docs#

I'll delete the old clustering code after the PR in the docs repo is merged.

PR #532, opened 2018-12-20
Removed legacy ML.NET clustering tutorial

pkulikov:
The new tutorial is already live: #521 and dotnet/docs#9620 are merged to live.
So, the old one can be removed.

/cc: @JRAlexander


JRAlexander:
Thanks, @pkulikov!

PR #535, opened 2018-12-20
Create samples for C# 6 that leverage the interactive window.

BillWagner:
Add new samples for object and collection initializers. These are for dotnet/docs#5169

Create samples for the new interactive "Explore C# 6" tutorial. This contributes to dotnet/docs#8206

### DIFF
--- a/snippets/csharp/new-in-6/initializers.cs
+++ b/snippets/csharp/new-in-6/initializers.cs
@@ -4,14 +4,14 @@ namespace Initializers
 {
     public class Examples
     {
-        // <ListInitializer>
+        // <CollectionInitializer>
         private Dictionary<int, string> messages = new Dictionary<int, string>
         {
             { 404, "Page not Found"},
             { 302, "Page moved, but left a forwarding address."},
             { 500, "The web server can't come out to play today."}
         };
-        // </ListInitializer>
+        // </CollectionInitializer>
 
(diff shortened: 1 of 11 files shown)


Thraka:
If these are written for C# 6, how do we make sure that the snippet isn't updated in the future and uses a feature from C# 7 or 8? It might be good to put the `csharp6` in the file path or something?

BillWagner:
Good catch @Thraka 

I meant to add the `<LangVersion>6.0</LangVersion>` in the csproj file.

Thraka:
Whoops, I should have said, I meant the files like: `classes-and-structs/object-collection-initializers/BasicObjectInitializers.cs` The path to these files is generic and likely represents the latest-and-greatest ? So having code in those files for the language specific articles could be troublesome in the future.

snippets/csharp/new-in-6/initializers.cs
```diff
@@ -5,11 +5,11 @@ namespace Initializers
     public class Examples
     {
         // <ListInitializer>
```
Thraka: This snippet tag says ListInitializer bu t it's using an IndexInitializer ?

BillWagner: Changed here, and in the use in dotnet/docs#9551

PR #543, opened 2018-12-27
Consolidating samples for how-to-write-text-to-a-file

v-thepet:
## Summary

Make the code examples in https://review.docs.microsoft.com/en-us/dotnet/standard/io/how-to-write-text-to-a-file?branch=pr-en-us-9760 more usable by providing an overall sample that can be run. 

Related to https://mseng.visualstudio.com/TechnicalContent/_workitems/edit/1385012.

### DIFF
--- a/snippets/csharp/VS_Snippets_CLR/conceptual.basicio.textfiles/cs/source.cs
+++ b/snippets/csharp/VS_Snippets_CLR/conceptual.basicio.textfiles/cs/source.cs
@@ -8,30 +8,29 @@ class WriteTextFiles
     static void Main(string[] args)
     {
 
-        // Example 1: shows how to write synchronously
+        // Example 1: Shows how to write synchronously
         // to a text file using StreamWriter
         // line by line
         WriteLineByLine();
 
-        // Example 2: Appending text to
+        // Example 2: Shows how to append text to
         // an existing file using StreamWriter
         AppendTextSW();
 
-        // Example 3: shows how to write a 
+        // Example 3: Shows how to write  
         // a simple string asynchronously
         // to a text file using StreamWriter
         string text = "This is a sentence.";
         WriteTextAsync(text);
 
-        // Example 4: shows how to write synchronously
+        // Example 4: Shows how to write synchronously
         // to a text file using File and then
-        // adding additional lines 
+        // add additional lines 
         WriteFile();
     }
 
     static void WriteLineByLine()
     {
-        // <SnippetWriteLine>
         // Create a string array with the lines of text
         string[] lines = { "First line", "Second line", "Third line" };
 
@@ -44,13 +43,11 @@ class WriteTextFiles
             foreach (string line in lines)
                 outputFile.WriteLine(line);
         }
-        // </SnippetWriteLine>
 
     }
 
     static void AppendTextSW()
     {
-        // <SnippetAppendText>
         // Set a variable to the My Documents path.
         string mydocpath = Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments);
 
@@ -58,11 +55,9 @@ class WriteTextFiles
         using (StreamWriter outputFile = new StreamWriter(Path.Combine(mydocpath,"WriteLines.txt"), true)) {
             outputFile.WriteLine("Fourth Line");
         }
-        // </SnippetAppendText>
 
     }
 
-    // <SnippetWriteAsync>
     static async void WriteTextAsync(string text)
     {
         // Set a variable to the My Documents path.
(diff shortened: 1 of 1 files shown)


v-thepet:
#hold-off - Please don't merge till I can consult with content owner about process.

snippets/csharp/VS_Snippets_CLR/conceptual.basicio.textfiles/cs/source.cs
```diff
@@ -8,30 +8,29 @@ class WriteTextFiles
...
         // to a text file using File and then
-        // adding additional lines 
+        // add additional lines 
         WriteFile();
     }
 
+    // Example 1: Write line by line to a new text file with StreamWriter
     static void WriteLineByLine()
     {
-        // <SnippetWriteLine>
```
BillWagner: These markers should stay.  They indicate blocks of code that are included in the corresponding article.

This comment applies to all the following comments that contain the `<Snippet*>` and `</Snippet*>` text.

v-thepet:
@BillWagner @rpetrusha - Thanks, I've put back the Snippet markers. I've changed the doc to run the whole sample so that it is runnable.

snippets/csharp/VS_Snippets_CLR/conceptual.basicio.textfiles/cs/source.cs
```diff
@@ -8,30 +8,31 @@ class WriteTextFiles
     static void Main(string[] args)
     {
 
-        // Example 1: shows how to write synchronously
+        // Example 1: Shows how to write synchronously
```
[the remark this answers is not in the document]
v-thepet: @rpetrusha - In the current live version of the article, there are 4 separate code samples (the 4 snippets from the original sample). They do not have any using statements, class declaration, Main method, so they will not run on their own. All the other samples in this set of articles run on their own, so it seemed confusing not to be able to run these 4 standalone. So there were 2 choices: create 4 separate, runnable samples (which I've now done), or present the entire sample as one and annotate it, which was what I originally did. I agree with you that the 4 samples are more usable. I am removing the numbered list from the topic (replacing it with the original 4 H2s), and most of the introductory code comments from the code, because they repeat the introductory text.

snippets/csharp/VS_Snippets_CLR/conceptual.basicio.textfiles/cs/append.cs
```diff
@@ -0,0 +1,27 @@
+using System;
+using System.IO;
+
+class Program
+{
+    static void Main(string[] args)
+    {
+        AppendText();
```
rpetrusha: There's no need here to put the code in a separate AppendText method that's called from Main; all of the code can just go in Main. That's true of the next examples as well.

snippets/csharp/VS_Snippets_CLR/conceptual.basicio.textfiles/cs/async.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+using System.IO;
+
+class Program
+{
+    static void Main(string[] args)
```
rpetrusha: Signature here would be `static async Task Main()`, and you'll need to add a `using` statement for hte System.Threading.Tasks namespace. In the text introducing the example, you should note that it requires C# 7.1 or later, where support for the async modifier on the entry point was added.

v-thepet: Yes, it turns out I was using an older version of C#, so could not put the async on Main. Creating the separate method and putting the async modifier on that made it run. (Then I added named methods to all the other samples for symmetry, but I agree that was uncalled-for.)  
With the updated version of #C, I can put the async on Main, and if I add the System.Threading.Tasks, the code works. I do wonder if most people will have C# => 7.1, because 7.0 will be the default (latest major version).

snippets/csharp/VS_Snippets_CLR/conceptual.basicio.textfiles/cs/file.cs
```diff
@@ -0,0 +1,23 @@
+using System;
+using System.IO;
+
+class Program
+{
+    static void Main(string[] args)
+    {
+        // Create a string array with the lines of text
+        string text = "First line" + Environment.NewLine;
```
pkulikov: @v-thepet comment doesn't match the code: not a string array is created.

Though, I wonder how much the comments are necessary: the code is self-documenting and additional lines make grasping it slower. Moreover, code comments are not localized.

rpetrusha: Thanks for noticing the incorrect comment, @pkulikov. I've updated the comment so that it's now accurate, @v-thepet , though I think that it's worthwhile to consider whether the comments are useful at all, especially since they aren't localized.  I'm going to merge this PR so that dotnet/docs#9760 can be built and reviewed. But in reviewing it, we should decide whether the comments should be removed and possibly incorporated into the descriptive text.

rpetrusha:
@BillWagner, could you approve this PR, please?

PR #552, opened 2019-01-05
Change to improve readability

volkan-yazilim:
## Summary

Describe your changes here.


BillWagner:
@volkan-yazilim I looked at this, and I need more information on your thinking before I proceed.

The original sample used the base type for the variables to illustrate how overridden methods use the runtime type instead of the compile time type. Your change does make it more readable, but obscures the purpose of the sample. That left me confused.

volkan-yazilim:
You are right. It does not add value.

PR #553, opened 2019-01-05
Make it clear that overridden methods may be overriden again

volkan-yazilim:
To make it clear why the 'override' word used in "virtual, abstract, or override" phrases in the text.

## Summary

Describe your changes here.


BillWagner:
I like the idea of adding one more layer of derived classes to this example to show how an overridden member can be overridden again. However, this example doesn't show that well. There's no good reason for this extra derived class to exist. If you can come up with a class derived from one of the existing `Shape` derived classes where the `Area` method does a different calculation, that would be a better example.

volkan-yazilim:
?

BillWagner:
closing due to the lack of a response.

@volkan-yazilim If you want to continue working on this, you can reopen this PR, and tag me for a new review.

PR #564, opened 2019-01-11
Minor code revisions for determining installed Framework versions

rpetrusha:
## Minor code revisions for determining installed Framework versions

Contributes to dotnet/docs#3991


rpetrusha:
Yes, it would be better, @BillWagner. I've made the changes in this PR. Thanks for going over the code so carefully.

PR #584, opened 2019-01-19
Rename method name

Potapy4:
## Summary

Refactoring for dotnet/docs#9835

### DIFF
--- a/snippets/csharp/VS_Snippets_VBCSharp/csrefKeywordsTypes/CS/keywordsTypes.cs
+++ b/snippets/csharp/VS_Snippets_VBCSharp/csrefKeywordsTypes/CS/keywordsTypes.cs
@@ -190,7 +190,7 @@ namespace csrefKeywordsMethodParams
     class DelegateTest
     {
         // Regular method that matches signature:
-        static double Foo(double input)
+        static double DoubleAmount(double input)
         {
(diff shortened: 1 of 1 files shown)


snippets/csharp/VS_Snippets_VBCSharp/csrefKeywordsTypes/CS/keywordsTypes.cs
```diff
@@ -190,15 +190,15 @@ static void Main()
     class DelegateTest
     {
         // Regular method that matches signature:
-        static double Double(double input)
+        static double Foo(double input)
```
BillWagner: The original name matched the functionality of the method better, although I can see the confusion from using a method name that matches a numeric type name.

I think we should pick a different name. How about `DoubleAmount` or `TwiceTheValue`. If you have a different idea, that would be great as well.

Potapy4: Well, let's use `DoubleAmount` instead of `Foo` as the method name

BillWagner:
Thanks for making the updates @Potapy4 

I'll approve this and :shipit: now. 

Thanks for making your first contribution to docs. 🎆

PR #605, opened 2019-01-30
Add unloadability sample

janvorli:
## Summary

The sample shows how to load an Assembly into an unloadable
AssemblyLoadContext, execute some methods in it and then unload it.

It also demonstrates usage of interface assembly shared between the host
and the assembly loaded into the unloadable AssemblyLoadContext that
simplifies calling methods over the boundary (without that, all calls
would have to use reflection).

And it also shows how to use the Unloading event to cleanup stuff loaded
into the AssemblyLoadContext that would otherwise prevent successful
unloading.


core/tutorials/Unloading/Plugin/PluginClass.cs
```diff
@@ -0,0 +1,48 @@
...
+        static GCHandle pluginHandle;
+
+        public static Interface GetInterface()
+        {
+            PluginClass plugin = new PluginClass();
+
+            // We create the GC handle just to demonstrate how we can do necessary cleanup before
+            // unloading. Creating a normal GC handle for the plugin instance would prevent unloading
+            // from succeeding.
+            pluginHandle = GCHandle.Alloc(plugin);
```
jkotas: What most people do with the same is to copy&paste into their project. I do not think we want to create the GCHandle for "demonstration purposes".

AaronRobinsonMSFT: Perhaps the callback could simply write to stdout - `"Clean up here"`?

janvorli:
@jkotas, @AaronRobinsonMSFT, I've updated the sample based on your feedback.

core/tutorials/Unloading/Plugin/PluginClass.cs
```diff
@@ -7,19 +7,16 @@ namespace Plugin
...
-            // unloading. Creating a normal GC handle for the plugin instance would prevent unloading
-            // from succeeding.
-            pluginHandle = GCHandle.Alloc(plugin);
-
-            // So we register handler for the Unloading event of the context that we are running in and 
-            // free the GC handle in there.
+            // We register handler for the Unloading event of the context that we are running in 
+            // so that we can perform cleanup of stuff that would otherwise prevent unloading
+            // (Like freeing GCHandles for objects of types loaded into the unloadable AssemblyLoadContext,
+            // terminating threads running code in assemblies loaded into the unloadable AssemblyLoadContext,
```
AaronRobinsonMSFT: Terminating threads is also a good one. You could start of a work thread here and then in the clean callback set an event and call `Join()` if you were looking for a more explicit clean up scenario. Other than that LGTM.

core/tutorials/Unloading/Host/Program.cs
```diff
@@ -0,0 +1,83 @@
...
+
+            // This initiates the unload of the HostAssemblyLoadContext. The actual unloading doesn't happen
+            // right away, GC has to kick in later to collect all the stuff.
+            alc.Unload();
+        }
+
+        static void Main(string[] args)
+        {
+            WeakReference hostAlcWeakRef;
+            ExecuteAndUnload(Path.GetFullPath("Plugin.dll"), out hostAlcWeakRef);
```
jkotas: Since this is loading "Plugin" built as regular csproj, would it make sense to use the @vitek-karas plugin loading APIs here to have a more complete sample for loading plugins?

jkotas: I am thinking we may want to have two samples - that match the main scenarios from the design doc:
- Plugin loading (this one)
- Dynamic compilation (compile a piece of code in memory with Roslyn, load it, run it and unload it)

janvorli: This makes sense. Let me update this PR by adding the usage of AssemblyDependencyResolver to resolve the plugin dependencies and load them into the context instead of loading just the plugin.dll itself into it.
I will create the Dynamic compilation sample as a separate PR later.

janvorli:
@jkotas I've updated the sample as you've suggested.

core/tutorials/Unloading/Plugin/Plugin.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Interface\Interface.csproj">
+      <Private>false</Private>
```
vitek-karas: Add a comment describing that this is the trick which makes the Interface.dll not present in the plugins directory. The name of the MSBuild property is misleading (to say the least).

janvorli: Added, thanks for the suggestion.

PR #619, opened 2019-02-04
Fix SnippetCreateMLContext

JRAlexander:
Fix SnippetCreateMLContext


JRAlexander:
Thanks, @luisquintanilla!

PR #620, opened 2019-02-04
Fix SnippetOutputPrediction

JRAlexander:
Fix SnippetOutputPrediction


JRAlexander:
Thanks, @luisquintanilla!

PR #630, opened 2019-02-08
sample code for nullable migration tutorial

BillWagner:
Contributes to dotnet/docs#8209

These samples show an existing application and walks the user through migrating it to make use of nullable reference types.

/cc @Jskeet


csharp/tutorials/nullable-reference-migration/finished/SimpleFeedReader.Tests/Services/NewsServiceTests.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Net;
+using System.Threading.Tasks;
+using AutoMapper;
+using SimpleFeedReader.Services;
+using SimpleFeedReader.ViewModels;
+using Xunit;
+
+namespace SimpleFeedReader.Tests.Services
+{
+#nullable enable
```
jskeet: I'd expect that within a genuinely *complete* migration, nullability would be enabled at a project level instead.

I appreciate that it's a lot of work to maintain three versions, but I wonder whether a model of before/during/after might be useful. I would imagine it would be *particularly* helpful to look at specific examples of what happens when you've migrated one piece of code but not others... and give some guidance about what to migrate first.

csharp/tutorials/nullable-reference-migration/finished/SimpleFeedReader/Pages/Error.cshtml.cs
```diff
@@ -0,0 +1,25 @@
...
+        public string? RequestId { get; set; }
+
+        public bool ShowRequestId => !string.IsNullOrEmpty(RequestId);
+
+        public void OnGet()
+        {
+            RequestId = Activity.Current?.Id ?? HttpContext.TraceIdentifier;
+        }
+    }
+#nullable restore
```
jskeet: Do we genuinely expect people to use `#nullable restore` much? From a quick scan of the code, it looks like you're only doing it in places where it wouldn't make any difference anyway.
I suspect someone migrating file-by-file will just use `#nullable enable` at the top of the file and leave it at that. If there's no reason to discourage that (and there may be something I haven't thought of!) it's probably worth explicitly mentioning that - and that `#nullable restore` is only really needed when you want to migrate *part* of a file.

csharp/tutorials/nullable-reference-migration/finished/SimpleFeedReader.Tests/Services/NewsServiceTests.cs
BillWagner: Good points. I'm going to defer these for the moment. I am sure that I will create a longer form tutorial that encompasses multiple projects (including test projects) as the release gets closer.

I need to migrate a few more codebases myself before I feel confident that I'm providing solid guidance on the strategy. (And I want to gather input from others that have migrated code).

csharp/tutorials/nullable-reference-migration/finished/SimpleFeedReader/Pages/Error.cshtml.cs
BillWagner: Right. I'm leaning toward pairing enable / restore because it does the right thing even when a file has multiple classes. I think it might also help if a class contains nested types. 

I agree that the most common use case is turning it on at the beginning of a file and leaving it on.  I'll likely revisit this first tutorial once more have been written.

PR #641, opened 2019-02-12
Add demo for Visio automation in .NET Core 3.0

JohnGoldsmith:
## Summary

This PR adds a demo project for Visio automation from .NET Core 3.0.  This is really a Visio version of the  Excel demo found here: https://github.com/dotnet/samples/tree/master/core/extensions/ExcelDemo

Fixes dotnet/docs#10323


core/extensions/VisioDemo/VisioDemo.sln
```diff
@@ -0,0 +1,25 @@
+﻿
```
AaronRobinsonMSFT: I would remove the SLN file. They are more annoying than helpful most of the time. In this with only a single project they definitely aren't needed.

JohnGoldsmith: That's fine.  Have added commit to remove the file.

PR #680, opened 2019-02-28
Update dataset usage and to version 0.11 for Sentiment Analysis 

JRAlexander:
Revising with UCI Sentiment Dataset for greater model accuracy, and updating to 0.11.0.

### DIFF
--- a/machine-learning/tutorials/SentimentAnalysis/Program.cs
+++ b/machine-learning/tutorials/SentimentAnalysis/Program.cs
@@ -51,24 +51,8 @@ namespace SentimentAnalysis
             Console.WriteLine("=============== End of process ===============");
         }
 
-        private static (IDataView trainSet, IDataView testSet) LoadAndTransformData(MLContext mlContext)
+        private static (IDataView trainSet, IDataView testSet) LoadData(MLContext mlContext)
         {
-            // The TextLoader loads a dataset with comments and corresponding postive or negative sentiment. 
-            // When you create a loader, you specify the schema by passing a class to the loader containing
-            // all the column names and their types. This is used to create the model, and train it. 
-            // Initialize our TextLoader
-           // TODO: Change to ReadFromTextFile
-            // <SnippetCreateTextLoader>
-            //return mlContext.Data.CreateTextLoader(
-            //    columns: new TextLoader.Column[]
-            //    {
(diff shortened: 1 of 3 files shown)


prathyusha12345:
@JRAlexander I tested the PR.working fine. But I have questions on the output. 
1.The accuracy metrics is 79%. - I believe accuracy should be more than 85%.Correct me if I am wrong.
2. See the probabilities in below image. how come a text with probability <50% is showing prediction correctly?
image

I believe that probability also should be > 50% to get prediction correctly. Correct me if I am wrong.

@CESARDELATORRE Adding Cesar for reference.

JRAlexander:
It leaves the door open for improving the model. This is the dataset we are using. It has better accuracy than using wikidetox.

CESARDELATORRE:
Well, a rule of thumb is to have an accuracy higher than 80%. However, depending on the business/domain it might need to be a lot higher or even be acceptable to be lower.

When using small datasets it is normal in most cases that the accuracy won't be very super high, neither the probabilities.

Since we're using small datasets for short trainings, I'd suggest not to show the probability in the code or output..

JRAlexander:
I'm going to leave it in there with an explanation about the small data

machine-learning/tutorials/SentimentAnalysis/Program.cs
```diff
@@ -134,117 +128,120 @@ public static void Evaluate(MLContext mlContext, ITransformer model)
...
                 {
-                    SentimentText = "I love this article."
+                    SentimentText = "I love this spaghetti."
                 }
             };
-            // </Snippet26>
+            // </SnippetCreateTestIssues>
 
-            // <Snippet27>
+            // <SnippetLoadModel>
```
luisquintanilla: For consistency with SaveModelAsFile, consider moving this logic into its own method.

JRAlexander: Ok, that will be a good rev. I'll add an issue to track it. Would rather go ahead and get it out.

luisquintanilla: Sounds good.

JRAlexander:
Added your suggestion as issue here https://github.com/dotnet/docs/issues/11221

JRAlexander:
Thanks, @luisquintanilla!

PR #699, opened 2019-03-08
Fixed a typo-bug that caused the InitializationSample in "Object and Collection Initializers" to throw a compiler error

jmarkman:
## Summary

I changed the line `Cat sameCat = new Cat("Fluffy"){ Age = 10 };` to take full advantage of object initialization and made spacing in the snippet consistent.

Fixes dotnet/docs#11223

### DIFF
--- a/snippets/csharp/programming-guide/classes-and-structs/object-collection-initializers/BasicObjectInitializers.cs
+++ b/snippets/csharp/programming-guide/classes-and-structs/object-collection-initializers/BasicObjectInitializers.cs
@@ -129,12 +129,19 @@ namespace object_collection_initializers
             // Auto-implemented properties.
             public int Age { get; set; }
             public string Name { get; set; }
+
+            public Cat() { }
+
+            public Cat(string name)
+            {
+                Name = name;
+            }
         }
 
         public static void Main()
         {
             Cat cat = new Cat { Age = 10, Name = "Fluffy" };
-            Cat sameCat = new Cat { Age = 10, Name = "Fluffy" };
+            Cat sameCat = new Cat("Fluffy"){ Age = 10 };
 
             List<Cat> cats = new List<Cat>
             {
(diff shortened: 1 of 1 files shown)


snippets/csharp/programming-guide/classes-and-structs/object-collection-initializers/BasicObjectInitializers.cs
```diff
@@ -134,19 +134,19 @@ public class Cat
         public static void Main()
         {
             Cat cat = new Cat { Age = 10, Name = "Fluffy" };
-            Cat sameCat = new Cat("Fluffy"){ Age = 10 };
+            Cat sameCat = new Cat { Age = 10, Name = "Fluffy" };
```
pkulikov: @jmarkman now, the updated line duplicates the line above. That looks strange.

I guess the intention was to show that one can mix using a constructor with the use of an object initializer. However, that is already covered in the earlier section in the same article:
https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/object-and-collection-initializers#object-initializers

So, removing line at all with the `sameCat` variable seems to be a better solution. What do you think?

jmarkman: Now that you've mentioned it, I would oppose that *and* my own submission. Instead, I'd propose that I should go back and simply change the `Cat` class to have a constructor with a name parameter. I was so focused on the fact that the line was erroring and that it was an initialization example that I didn't stop to think that it was mixing them together; couldn't see the forest for the trees.

pkulikov: > ...change the `Cat` class to have a constructor with a name parameter.

Agree, it's another possible solution. Then, the parameter-less constructor must be explicitly added to the class as well.

jmarkman: I've just added the parameterless and parameterized ctors to the `Cat` class and reverted my previous change to the hybrid constructor/initializer line.

PR #701, opened 2019-03-08
Updated event dictionary example

pkulikov:
Replaces dotnet/docs#9937 and dotnet/docs#9953. 
Simplifies the event dictionary example according to https://github.com/dotnet/docs/pull/9937#issuecomment-466465344.

/cc @Kevinlinpr
@BillWagner please review

@BillWagner the usage is as follows
```csharp
var p = new PropertyEventsSample();
p.Event1 += Delegate1Method;
p.Event1 += Delegate1Method;
p.Event1 -= Delegate1Method;
p.RaiseEvent1(2);
```

where `Delegate1Method` is the method group. Did you have that in mind in the following sentence:
>The example that calls this implementation should keep a reference to the delegates used to subscribe to events so they correctly unsubscribe.

### DIFF
--- a/snippets/csharp/VS_Snippets_VBCSharp/csProgGuideEvents/CS/Events.cs
+++ b/snippets/csharp/VS_Snippets_VBCSharp/csProgGuideEvents/CS/Events.cs
@@ -451,26 +451,26 @@ namespace WrapEvent
 //-----------------------------------------------------------------------------
 namespace WrapHash
 {
-    using System;
     //<Snippet9>
-    using System;    
-    using System.Collections.Generic;    
+    using System;
+    using System.Collections.Generic;
 
     public delegate void EventHandler1(int i);
-
     public delegate void EventHandler2(string s);
 
     public class PropertyEventsSample
     {
-        private readonly Dictionary<string, Delegate> _eventTable;
-        private readonly List<EventHandler1> _event1List = new List<EventHandler1>();
-        private readonly List<EventHandler2> _event2List = new List<EventHandler2>();
+        private const string Event1Key = nameof(Event1);
+        private const string Event2Key = nameof(Event2);
+
+        private readonly Dictionary<string, Delegate> handlers;
+
         public PropertyEventsSample()
         {
-            _eventTable = new Dictionary<string, Delegate>
+            handlers = new Dictionary<string, Delegate>
             {
-                {"Event1", null},
-                {"Event2", null}
+                [Event1Key] = null,
+                [Event2Key] = null
             };
         }
 
@@ -478,23 +478,16 @@ namespace WrapHash
         {
             add
             {
-                _event1List.Add(value);
-                lock (_eventTable)
+                lock (handlers)
                 {
-                    _eventTable["Event1"] = (EventHandler1) _eventTable["Event1"] + value;
+                    handlers[Event1Key] = (EventHandler1)handlers[Event1Key] + value;
                 }
             }
             remove
             {
-                if (!_event1List.Contains(value)) return;
-                _event1List.Remove(value);
-                lock (_eventTable)
+                lock (handlers)
                 {
-                    _eventTable["Event1"] = null;
-                    foreach (var event1 in _event1List)
-                    {
-                        _eventTable["Event1"] = (EventHandler1) _eventTable["Event1"] + event1;
-                    }
+                    handlers[Event1Key] = (EventHandler1)handlers[Event1Key] - value;
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


BillWagner:
I like this solution.  Before I approve it, I'd like @kevinlinpr to add thoughts as well.

That said, I think we should also make a small modification to the article where this sample is included.  This code correctly locks the dictionary only when the handlers are being accessed or modified. It's critically important that the dictionary is not locked when the event handlers are called. That could lead to deadlock. (Imagine an event handler that tried to remove itself).  I think it's important to point that out in the accompanying text.

pkulikov:
> Imagine an event handler that tried to remove itself

If it removes itself on the same thread as it has been invoked, that's fine as `lock` is re-entrant. That said, such a case is not a guarantee.

pkulikov:
@BillWagner I've created the PR with the article update: dotnet/docs#11254

Kevinlinpr:
First of all, Thank you so much,both of you two @BillWagner @pkulikov 
@pkulikov I'v reviewed your change,you made the code easy to read and understand, I really appreciate that.I think we can make operator ‘-’ more controllable through the solution which we subtract event one by one( to avid the warning about unpredictable subtraction).I agree with you to put this code in the event example.
@BillWagner **But the problem is still there, it is still unpredictable.**
pkulikov's change still stay the original place, what if I want to subtract a events list but this events list is not the subset of the total set?
```
Example: 
1.ABC - BC = A  It success because BC is subset of ABC
2.ABC - CB = ABC It failed because CB is not subset of ABC
3.ABC - CB question now we change into ABC - C - B or ABC - B - C .Then it will become predictable!This is what I say to make it controllable one by one.
```
**But let me guess the extreme case:**
I want to subtract some 'event' during **Runtime execution**,when the **amount** of events **become more and more huge**,I do not willing subtract it one by one repeat this easy way again and again. so we will always have to subtract a event list. 
By the way ,we are programmer we can avoid this unpredictable during coding,but it is **predictable only before the compile checking**. In runtime execution ,If I have a event list pool to collect the event which we will **subtract all of them in only one times**?So we need a more smart solution to face this extreme case.
**Maybe it is ok before compile,maybe not during runtime execution.**
How about your @pkulikov @BillWagner  idea?

BillWagner:
@Kevinlinpr Your analysis is correct, but I think these more extreme cases are best handled not in the code but in the accompanying article.

The code in this PR is correct as long as the caller adds and removes the same delegate objects. 

Your failure cases all rely on the fact that the argument to `add` or `remove` could be a delegate, or a multicast delegate (which has a collection of delegates.)  A caller could add `A`, then add `B`, and try to remove a multicast delegate that contains both `A` and `B`.

If this were to try to solve that problem, I would code the `add` method to do the following:
1. If the object being added is a `MulticastDelegate`, get its invocation list, and add each individual delegate to the list of event handlers.
1. Otherwise, add the single delegate to the list.

Then, `remove`  would perform something similar:
1. If the object being removed is a `MulticastDelegate`, remove each individual delegate from from the list of event handlers.
1. Otherwise, remove the single delegate from the list.

But wait, that's not sufficient for all cases: any of the delegates on the invocation list could be `MulticastDelegate` objects. Those would contain their own invocation list. We're now forced to create a recursive algorithm for both `add` and `remove`.

The most common use case is the one in this sample: adding and removing the same object from the invocation list.  For that reason, I'd prefer demonstrating that use case, then explain its limitations in the accompanying text.

Thoughts?

pkulikov:
@Kevinlinpr @BillWagner 

As far as I understand, the containing article suggests a memory optimization for the following, let's call it simple, implementation:
```csharp
public class SimpleEvent
{
    public event EventHandler1 Event1;
    public event EventHandler2 Event2;

    internal void RaiseEvent1(int i) => Event1?.Invoke(i);
    internal void RaiseEvent2(string s) => Event2?.Invoke(s);
}
```

That is, the simple implementation and the dictionary implementation *must* have the same behavior. The simple implementation uses the standard `-` operator to unsubscribe. That's why, I don't think we can change the semantics of this operation in the dictionary implementation.

As for predictability, it looks that the "predictability" notion is subjective. The behavior of the `-` operator for delegate operands is described in the [language specification (the last point in the linked section)](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#subtraction-operator). According to the spec, `ABC - CB` results in `ABC`. That's why, to me, such behavior is predictable. Maybe, not intuitive, but predictable. I would be surprised if `ABC - CB` would result in `A`. With that, I think the [reference article about `-` operator](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/subtraction-operator) should provide more information about delegate removal (occasionally, I'll revise that article).

I understand that the behavior suggested by @Kevinlinpr might be desired one. However, I think it's a user-specific requirement, which differs from the standard behavior. Also, I doubt that Microsoft docs should call the behavior according to the spec unpredictable :).

Kevinlinpr:
@BillWagner Thanks for considering my analysis. Do you know why I pay more attention to this unpredictable extreme case? I learn the C# programming language standard grammar on this Doc, which must be the best comprehensive documents. But after that, I try to do some multicast delegate subtraction. The results of execution is not my prediction.** The only way to avoid logical error is that improve it in code level, not the user-warning level by using text comments.** 
For the common use,your preservation is good.
But you need to explain its limitations in a new page and give a redirect url to let user know the solution about this unpredictable extreme case.
In the new page, please code the add method to do the following:
1. If the object being added is a MulticastDelegate , get its invocation list, and add each individual delegate to the list of event handlers.
2. Otherwise, add the single delegate to the list.

Then, remove would perform something similar:
1. If the object being removed is a MulticastDelegate , remove each individual delegate from from the list of event handlers.
2. Otherwise, remove the single delegate from the list.
This new change will make the add and remove results (both multicast delegate and single delegate) predictable not only before the compile build but also in Runtime execution.
Thank you so much.

Kevinlinpr:
@pkulikov yeah I think that it is very necessary to explain why it is unpredictable and how we can avoid not only before compile build but also in runtime execution in a new page and preserve this solution for the common use.
Thank you so much for your attention.

Thraka:
This looks like a good rewrite. Do you not need to check if the item exists in the dictionary like the old code was doing?

BillWagner:
The behavior from this PR is not "unpredictable". It does match the behavior in the [C# language spec](https://docs.microsoft.com/dotnet/csharp/language-reference/language-specification/delegates#delegate-invocation). The spec has several examples that show how delegates and events behave when you add or remove them. It may be surprising if you aren't familiar with the specification and the behavior defined there.

A different sample that creates different semantics for delegate addition and removal would be more surprising for those that would expect the behavior to match the spec.

It would help to add the link I added above to explain why the behavior was chosen. (I've made that comment on dotnet/docs#11254). I don't think adding a different sample that creates a non-conforming pattern of delegate addition and subtraction would help understanding.

PR #713, opened 2019-03-14
Sample output from 2007 to 2006

Luminca:
## Summary

The ConvertTimeToUtc(DateTime) example contains a typo #1693
Changed the sample output shown for the third conversion from 2007 to 2006.


Luminca:
Thank you :)

PR #722, opened 2019-03-17
:wastebasket: Removed not used snippets

pkulikov:
After dotnet/docs#11313 is merged, a number of snippets is not referenced anymore and can be removed.
PR #716 added replacement snippets for the lambda expressions article.


BillWagner:
Adding the 🗑 emoji in the title in case I need to find these changes later on a build warning.

PR #728, opened 2019-03-19
responding to feedback.

BillWagner:
1. typo "mepty"
1. copy / paste error on error writeline.
1. remove unnecessary `when` clause.


BillWagner:
Thanks @rpetrusha 

The "unnecessary when clause" was only in the interim code, so that change is only in dotnet/docs#11392

PR #741, opened 2019-03-23
Fix compile

lindexi:
## Summary

The bracket can not be added after the class name.

[Assembly.GetCallingAssembly Method (System.Reflection)](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.getcallingassembly?view=netframework-4.7.2#System_Reflection_Assembly_GetCallingAssembly )


lindexi:
The core change is https://github.com/dotnet/samples/pull/741/files#diff-f6b89c0100c26208f0a398042779d5d3L26

PR #772, opened 2019-04-01
Revisions for Reiwa era

rpetrusha:
## Revisions for Reiwa era

@tarekgh

### DIFF
--- a/snippets/csharp/VS_Snippets_CLR/conceptual.calendars/cs/instantiatewithera1.cs
+++ b/snippets/csharp/VS_Snippets_CLR/conceptual.calendars/cs/instantiatewithera1.cs
@@ -25,13 +25,3 @@ public class Example
       }                        
    }
 }
-// The example displays the following output:
-//    Date instantiated without an era:
-//    1/1/2 in Japanese Calendar -> 1/1/2020 in Gregorian
-//    
-//    Dates instantiated with eras:
-//    1/1/2 era 5 in Japanese Calendar -> 1/1/2020 in Gregorian
-//    1/1/2 era 4 in Japanese Calendar -> 1/1/1990 in Gregorian
-//    1/1/2 era 3 in Japanese Calendar -> 1/1/1927 in Gregorian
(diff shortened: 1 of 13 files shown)


snippets/csharp/VS_Snippets_CLR/conceptual.calendars/cs/instantiatewithera1.cs
```diff
@@ -28,11 +27,11 @@ public static void Main()
 }
 // The example displays the following output:
 //    Date instantiated without an era:
-//    1/1/2 in Japanese Calendar -> 1/1/1990 in Gregorian
```
tarekgh: The change here looks good but it depends on the updates of the era data and also we'll need to change this doc again in the future when getting any new era. 

Can we just change the part at the end to tell something can explain it? I mean instead of having 

// 1/1/2 in Japanese Calendar -> 1/1/2020 in Gregorian

maybe we can say, the code sample will show the equivalent Gregorian date that match year 2 month 1 day 1 of the current era on the system.

tarekgh:
> 

I would avoid showing the exact output of this code because it will need to change again in the future, and also the code depends on the date pattern "d" which can be easily changed on the system and produce different output.

---
Refers to: snippets/standard/datetime/calendars/current-era/cs/Program.cs:22 in c61e2ce. [](commit_id = c61e2ceeeecb27488edfd1699508acd8ecfcbc94, deletion_comment = False)

tarekgh:
I would say, we need to change all doc sample here to not mention the exact output but instead explain what to expect the output will look like. the reasons are: 
- Showing the output will depends on the machine settings (e.g. the machine has the era updates and what date format settings).
- We'll have to revisit all these samples again every time we have a new era.
- The output possibly can be different on Linux machines depending on the ICU version we use there.

rpetrusha:
@tarekgh, I've removed output except for the leap year/day examples, since leap years/days have historically caused confusion. In all other cases, I've modified the text to describe the output. I'll merge this, but please feel free to review these changes (as well as the changes in the related PRs) and let me know if you'd like anything further to be changed.

PR #779, opened 2019-04-02
Corrected lower bounds of VB for loop

rpetrusha:
## Corrected lower bounds of VB for loop

Related to #775, #776


rpetrusha:
We all need to make the same mental note, @BillWagner.  I'm certain I've missed as many or more than you have.

PR #787, opened 2019-04-04
Added snippets for the logical operators overview

pkulikov:
Supports dotnet/docs#11656

### DIFF
--- a/snippets/csharp/language-reference/operators/LogicalOperators.cs
+++ b/snippets/csharp/language-reference/operators/LogicalOperators.cs
@@ -1,0 +1,165 @@
+using System;
+
+namespace operators
+{
+    public static class LogicalOperators
+    {
+        public static void Examples()
+        {
+            Console.WriteLine("==== ! operator");
+            Negation();
+
+            Console.WriteLine("==== &, |, ^ operators");
+            And();
+            Or();
+            Xor();
+
+            Console.WriteLine("==== && and || operators");
+            ConditionalAnd();
+            ConditionalOr();
+
+            Console.WriteLine("==== Compound assignment and precedence");
+            CompoundAssignment();
+            Precedence();
+        }
+
+        private static void Negation()
+        {
+            // <SnippetNegation>
+            bool passed = false;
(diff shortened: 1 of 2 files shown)


snippets/csharp/language-reference/operators/LogicalOperators.cs
```diff
@@ -0,0 +1,165 @@
...
+
+            Console.WriteLine("==== Compound assignment and precedence");
+            CompoundAssignment();
+            Precedence();
+        }
+
+        private static void Negation()
+        {
+            // <SnippetNegation>
+            bool passed = false;
```
BillWagner: This comment applies to all the samples, but I'll make it just this once.

Do you think adding examples that include `bool?` operands would be useful? 

It does increase the complexity, but it would fill out the information in [this table](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#nullable-boolean-logical-operators) in the spec.

Thoughts?

pkulikov: @BillWagner just including nullable types into the samples might be misleading, because it might be interpreted as the example of the [operator lifting](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/nullable-types/using-nullable-types#operators):
>The predefined unary and binary operators and any user-defined operators that exist for value types may also be used by nullable types.

While the referenced table shows that both `&` and `|` operators are not lifted versions of `&` and `|` for `bool` operands. So, it would be useful to pronounce it explicitly. The docs already include the section with that table:
https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/nullable-types/using-nullable-types#the-bool-type

I suggest to **move** this section to the new overview of logical operators and rename it to "Nullable Boolean logical operators" (as in the language spec). 

What do you think? (If the full table is in the article, I don't think additional samples are necessary.)

BillWagner: I really like that plan @pkulikov 

Let's do it!

PR #788, opened 2019-04-04
Fix data marshaling

NextTurn:
Continued from dotnet/docs#11450

It's a common mistake to marshal pointers as `int` type.

Related to dotnet/csharplang#2324. Is there anyone to review the PRs there?

### DIFF
--- a/snippets/csharp/VS_Snippets_CLR/conceptual.interop.marshaling/cs/msgbox.cs
+++ b/snippets/csharp/VS_Snippets_CLR/conceptual.interop.marshaling/cs/msgbox.cs
@@ -9,20 +9,20 @@ public class LibWrap
     [DllImport("User32.dll", EntryPoint = "MessageBox",
         CharSet = CharSet.Auto)]
     public static extern int MsgBox(
-        IntPtr hWnd, string text, string caption, uint type);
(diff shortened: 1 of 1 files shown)


snippets/csharp/VS_Snippets_CLR/conceptual.interop.marshaling/cs/msgbox.cs
```diff
@@ -9,20 +9,20 @@ public class LibWrap
     [DllImport("User32.dll", EntryPoint = "MessageBox",
         CharSet = CharSet.Auto)]
     public static extern int MsgBox(
-        int hWnd, string text, string caption, uint type);
+        IntPtr hWnd, string text, string caption, uint type);
```
NextTurn: Should the parameters be renamed to `lpText`, `lpCaption` and `uType` to follow the naming conventions (as I did in dotnet/docs#11450)?

In https://docs.microsoft.com/dotnet/standard/native-interop/best-practices

> ✔️ DO use the same naming and capitalization for your methods and parameters as the native method you want to call.

BillWagner: Yes, that would be better. Thanks.

PR #794, opened 2019-04-05
Add comment with expected output for Uri.Equals

SeanKilleen:
## Summary
Fixes dotnet/dotnet-api-docs#1815

Per the issue, the C#, C++, and VB docs did not display what the expected output would be.

This updates the text with the correct output.


SeanKilleen:
(Just to make sure I wasn't crazy, I ran the sample in a console app and verified the output. 😆 )

PR #803, opened 2019-04-09
add generic pattern match sample

BillWagner:
Supports dotnet/docs#10334


snippets/csharp/language-reference/keywords/switch/type-pattern3.cs
```diff
@@ -0,0 +1,55 @@
...
+            default:
+                Console.WriteLine("Null passed to this method");
+                break;
+        }
+    }
+}
+// The example displays the following output:
+//     An array with 5 elements.
+//     4 items
+// </Snippet1>
```
rpetrusha:
```suggestion
//     Null passed to this method
// </Snippet1>
```

PR #815, opened 2019-04-13
Fix copy/paste bug in Roslyn SDK tutorial

codecadwallader:
##  Fix copy/paste bug in unit test data samples. 

There are unit test examples for integers and strings, but the string unit tests mirror the integer unit tests.  This fix simply gives the string unit tests the expected initialization values.

Fixes dotnet/docs#11826


codecadwallader:
Cool, thanks for reviewing it quickly! :)

PR #837, opened 2019-04-22
Remove "recursive" from comment

BillWagner:
Fixes #11744

The algorithm isn't really "recursive" because the `GetDirectories` will return all sub-directories, at any level from teh current directory. Therefore, the later calls to `Search` all search only for files.

Update the comment to better reflect that.

### DIFF
--- a/csharp/events/Program.cs
+++ b/csharp/events/Program.cs
@@ -96,7 +96,7 @@ namespace EventSampleCode
                 {
                     directoryChanged?.Invoke(this,
                         new SearchDirectoryArgs(dir, totalDirs, completedDirs++));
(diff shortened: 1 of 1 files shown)


csharp/events/Program.cs
```diff
@@ -96,7 +96,7 @@ public void Search(string directory, string searchPattern, bool searchSubDirs =
                 {
                     directoryChanged?.Invoke(this,
                         new SearchDirectoryArgs(dir, totalDirs, completedDirs++));
-                    // Recursively search this child directory:
+                    // Search each sub-directory for files that match the search pattern:
```
rpetrusha: Perhaps "Search 'dir' and its  subdirectories for files that match the search pattern."?

PR #839, opened 2019-04-22
Migrate Movie recommender to 1.0.0-preview

prathyusha12345:
Migrate Movie recommender to ML.Net 1.0.0-preview


prathyusha12345:
@JRAlexander  I see the commits on TaxiFare. I don't remember that I have taken this commit. Anyhow I checked now by merging master from upstream into the current branch and only 2 files are changed that are specific to Movie Recommender only. There are no issues like merge conflicts or something. 

Can you check once and let me know in case of changes

JRAlexander:
Thanks, @prathyusha12345!!

PR #841, opened 2019-04-23
Removed redundant variable

Zaph-x:
The variable 10 has been removed as it simply did not make sense to declare a variable. The variable 10 was redundant, as it would suffice to just add 10

## Summary

The variable `ten` as been removed in favor of a 10. The variable seemed redundant as it didn't actually change the behavior of the compiler. The comments have been updated to reflect this change.

This can be proven with the following code:
```csharp
int max = int.MaxValue;
Console.WriteLine(max); // 2147483647
Console.WriteLine(max + 10); // -2147483639
checked {
    int y = max + 10; // An exception is thrown
}
```


BillWagner:
@Zaph-x Thanks for creating this PR. In most instances, you are correct that the variable `ten` is redundant.

However, in this specific instance, the variable declaration is important.  The expression:

```csharp
2147483647 + 10
```

is evaluated by the compiler, and produces a compile-time error. The introduction of the `ten` variable ensures that the expression is evaluated at runtime, producing a runtime exception.

Therefore, I'm closing this PR.

PR #850, opened 2019-04-25
Clarify DateTime implicit conversion example

Crown0815:
## Summary
It took some time for me to understand why the last example has -08:00 as offset. Since the example seemed to be about the addition of a TimeSpan the change in offset was quite confusing. 

Therefore, I split the example into two parts, one only focusing on the addition of a TimeSpan and the second on the adaption of the daylight time specific time zone. 

I also fixed a typo in the year given in line 84.

Fixes dotnet/dotnet-api-docs#2355

### DIFF
--- a/snippets/csharp/VS_Snippets_CLR_System/system.DateTimeOffset.Operators/cs/Operators.cs
+++ b/snippets/csharp/VS_Snippets_CLR_System/system.DateTimeOffset.Operators/cs/Operators.cs
@@ -81,7 +81,7 @@ public class Class1
    {
       // <Snippet7>
       DateTimeOffset timeWithOffset; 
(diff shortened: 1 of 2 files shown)


BillWagner:
Thanks for making this update @Crown0815 The C# changes look great.

Can you make the corresponding update to the VisualBasic sample? The source is here: https://github.com/dotnet/samples/blob/master/snippets/visualbasic/VS_Snippets_CLR_System/system.DateTimeOffset.Operators/vb/Operators.vb#L113

After that, we'll merge this change.

/cc @rpetrusha

Crown0815:
Thank you very much for your feedback @BillWagner. 

I adopted my changes to the C# sample to the VisualBasic sample as well. Please let me know if any further adoptions are necessary. 

I am very grateful for the opportunity to contribute.

PR #854, opened 2019-04-27
Fix code format

lindexi:
In [corefx/coding-style.md at master · dotnet/corefx](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md ) NO.1, we use Allman style braces, where each brace begins on a new line. A single line statement block can go without braces but the block must be properly indented on its own line and must not be nested in other statement blocks that use braces (See issue 381 for examples). One exception is that a using statement is permitted to be nested within another using statement by starting on the following line at the same indentation level, even if the nested using contains a controlled block. 

NO.11, we use language keywords instead of BCL types.


lindexi:
[Span<T> Struct (System) | Microsoft Docs](https://docs.microsoft.com/en-us/dotnet/api/system.span-1?view=netstandard-2.1 )

image

PR #856, opened 2019-04-27
Move P/Invokes to NativeMethods class

NextTurn:
The same as dotnet/docs#11985


BillWagner:
@NextTurn I reviewed the code changes this morning.  Overall they look great.

I'm curious about your choice to declare the `NativeMethods` class as `internal`. It works here in this sample, but would it be the standard practice in a larger application?

NextTurn:
@BillWagner 

In [CA1060: Move P/Invokes to NativeMethods class](https://docs.microsoft.com/visualstudio/code-quality/ca1060-move-p-invokes-to-nativemethods-class#rule-description)

> These classes are declared as `internal` (`Friend`, in Visual Basic) and declare a private constructor to prevent new instances from being created. The methods in these classes should be `static` and `internal` (`Shared` and `Friend` in Visual Basic).

It seems to be considered as a good practice.

PR #865, opened 2019-05-02
Formatting examples for string best practices

rpetrusha:
## Formatting examples for string best practices

Contributes to dotnet/docs#11429

Related to dotnet/docs#12158

### DIFF
--- a/snippets/standard/base-types/string-practices/cs/tostring.cs
+++ b/snippets/standard/base-types/string-practices/cs/tostring.cs
@@ -5,14 +5,14 @@ class Program
 {
     static void Main(string[] args)
     {
-        // <Snippet2>
-        string value = "The amount is " + 126.03.ToString(CultureInfo.InvariantCulture) + ".";
-        Console.WriteLine(value);
-        // </Snippet2>
-
         // <Snippet1>
-        value = "The amount is " + 126.03 + ".";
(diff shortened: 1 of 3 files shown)


snippets/standard/base-types/string-practices/cs/tostring.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    static void Main(string[] args)
+    {
+        // <Snippet2>
+        string value = "The amount is " + 126.03.ToString(CultureInfo.InvariantCulture) + ".";
+        Console.WriteLine(value);
+        // </Snippet2>
+
+        // <Snippet1>
+        value = "The amount is " + 126.03 + ".";
```
pkulikov: In the article text, this snippet goes first. As a result, in the article the first snippet doesn't declare the `value` variable, but the second declares. Might be confusing.

I would use two different `string` variables for different snippets: 
```csharp
string messageInInvariantCulture = ...
// and
string messageInCurrentCulture =...
```

rpetrusha: Thanks for noticing that, @pkulikov. I've reversed the order and declared separate string variables named `concat1` and `concat1`.

BillWagner:
Updated base branch to master.

PR #876, opened 2019-05-05
Use string.IsNullOrEmpty

NextTurn:
Contributes to dotnet/dotnet-api-docs#1786

### DIFF
--- a/snippets/csharp/VS_Snippets_CLR_System/system.string.isnullorempty/cs/isnullorempty1.cs
+++ b/snippets/csharp/VS_Snippets_CLR_System/system.string.isnullorempty/cs/isnullorempty1.cs
@@ -14,7 +14,7 @@ public class Example
    {
       bool result;
       // <Snippet1>
-      result = string.IsNullOrEmpty(s);
+      result = s == null || s == string.Empty;
       // </Snippet1>
       return result;
    }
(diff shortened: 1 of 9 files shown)


snippets/csharp/VS_Snippets_WebNet/WebParts_ToolZone_Overview/CS/Toolzone_overview.cs
```diff
@@ -9,7 +9,7 @@ public partial class Toolzone_overview_cs : System.Web.UI.Page
 {
   protected void Button1_Click(Object sender, EventArgs e)
   {
-    if(TextBox1.Text == String.Empty | TextBox1.Text.Length < 0)
```
NextTurn: What's the hack?

snippets/csharp/VS_Snippets_CLR_System/system.string.isnullorempty/cs/isnullorempty1.cs
```diff
@@ -14,7 +14,7 @@ private static bool TestForNullOrEmpty(string s)
    {
       bool result;
       // <Snippet1>
-      result = s == null || s == String.Empty;
+      result = string.IsNullOrEmpty(s);
```
pkulikov: I guess this change should be reverted, because it's for this part of the doc:
image

https://docs.microsoft.com/en-us/dotnet/api/system.string.isnullorempty?view=netframework-4.8#remarks

snippets/csharp/VS_Snippets_WebNet/WebParts_ToolZone_Overview/CS/Toolzone_overview.cs
BillWagner: I'm at a loss as to what this was supposed to be. A quick search seems to indicate that it's not used at all.

I'll accept this PR with this change.

PR #883, opened 2019-05-07
Updated method name to use .NET naming guidelines

fr3gu:
## Summary

Changed method name from parse to Parse to use .NET naming guidelines


fr3gu:
Yay! Thanks!

PR #903, opened 2019-05-14
Formatting of try-catch

ThomasArdal:
Newline for curly braces.

Besides this, I'm not sure this example works. I've tried copying the code into a new console application and the debugger doesn't break inside the catch of the `AggregateException`. When awaiting a task like this, the original exception is thrown. In this case the `UnauthorizedAccessException`.

If doing the following, the `AggregateException` is thrown:

```
task1.Wait();
```

### DIFF
--- a/snippets/csharp/VS_Snippets_CLR_System/system.aggregateexception.class/cs/exception1.cs
+++ b/snippets/csharp/VS_Snippets_CLR_System/system.aggregateexception.class/cs/exception1.cs
@@ -17,10 +17,12 @@ class Example
 
       // Use this line to throw an exception that is not handled.
       // Task task1 = Task.Factory.StartNew(() => { throw new IndexOutOfRangeException(); } );
-      try {
+      try
+      {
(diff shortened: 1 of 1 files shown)


ThomasArdal:
@BillWagner What about my comment about the sample code not working (I think, but I'm not sure)?

BillWagner:
@ThomasArdal You are correct.

I'm guessing this sample was changed from using `Task.Wait` to using `await`, as part of general cleanup for best practices.

I think the best fix is to update the sample to show both catch clauses so readers can see the difference:

```csharp
try
{
    await task1;
   // task1.Wait();
} catch (IndexOutOfRangeException)
{
   Console.WriteLine("caught index of out range, thrown by awaiting the task");
}
catch (AggregateException ex)
{
   Console.WriteLine("caught aggregate exception, thrown by calling task.Wait");
   // etc.
}
```

ThomasArdal:
Internally would be fine :smile:

PR #913, opened 2019-05-17
Remove property on v1 of object

Thraka:
## Summary

Remove the property from the v1 object that is added in the v2 object.

Fixes dotnet/docs#10268

### DIFF
--- a/snippets/core/whats-new/whats-new-in-30/cs/Program.cs
+++ b/snippets/core/whats-new/whats-new-in-30/cs/Program.cs
@@ -102,9 +102,6 @@ namespace whats_new
                 var name = config.GetProperty("name").GetString();
                 Console.WriteLine($"Config: {name}");
             }
-
-            JsonPerson[] a;
-            a = { { new JsonPerson() } };
         }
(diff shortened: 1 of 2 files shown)


snippets/core/whats-new/whats-new-in-30/cs/Program.cs
```diff
@@ -102,6 +102,9 @@ public static void ReadJson(string jsonString)
                 var name = config.GetProperty("name").GetString();
                 Console.WriteLine($"Config: {name}");
             }
+
+            JsonPerson[] a;
+            a = { { new JsonPerson() } };
```
pkulikov: @Thraka that addition looks strange: declaration and initialization are separated; also `a` is not used in the snippet.

Thraka: Oh weird. I think this file snuck in this PR, it doesn't belong

Thraka: Thanks @pkulikov for finding that 😄

PR #929, opened 2019-05-27
Added snippets for the subtraction operator article

pkulikov:
Supports dotnet/docs#

@BillWagner those code examples are added into the new location, not the *snippets* folder. Is that a proper location?

### DIFF
--- a/csharp/language-reference/operators/Program.cs
+++ b/csharp/language-reference/operators/Program.cs
@@ -1,0 +1,14 @@
+﻿using System;
+
+namespace operators
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            Console.WriteLine("============== - operator examples =============");
+            SubtractionOperator.Examples();
+            Console.WriteLine();
+        }
+    }
+}
--- a/csharp/language-reference/operators/SubtractionOperator.cs
+++ b/csharp/language-reference/operators/SubtractionOperator.cs
@@ -1,0 +1,58 @@
+using System;
+
+namespace operators
+{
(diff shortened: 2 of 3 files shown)


BillWagner:
> those code examples are added into the new location, not the snippets folder. Is that a proper location?

Yes. That's perfect.

I really like the visual nature of these samples @pkulikov The visual nature of the sample is very compelling.

One global comment I have:  An area of confusion for many C# developers is that delegate removal operates on the *delegate object* not the method. And related is that removing a delegate that is not part of the multicast delegate does nothing, but isn't an error condition.

For example: something like this:

```csharp
    Action a = () => Console.Write("a");
    Action b = () => Console.Write("b");
    var printer = a + b + a;
    printer();  // output: aba
    Console.WriteLine();
    Action aPrime = () => Console.Write("a");
    printer = printer - aPrime;
    printer(); // output: aba
    Console.WriteLine();
    printer = printer - a;
    printer(); // output: ab
```

pkulikov:
@BillWagner that's a great remark, thanks! Definitely it should be covered (both in the sample and the article).

That said, I'll be able to make such an edition in a week. So, there are at least two ways to proceed: (1) to keep those two PRs open for a week, or (2) merge them and I'll submit dedicated PRs with the clarifications in a week. I'm fine with both ways.

BillWagner:
@pkulikov 

Let's make that update in another PR.  On my list this week is getting dotnet/docs#12244 merged, and then making some other TOC updates. Leaving this open for an entire week will mean more merge conflicts. The later changes should be fine without any merge isues.

PR #936, opened 2019-05-30
Update discard-pattern2.cs (Issue #8727)

psaxton:
## Summary
Update discard-pattern2.cs
Changed implementation to actually use the switch and discard language features being demonstrated.

Fixes dotnet/docs#8727 or dotnet/dotnet-api-docs#Issue_Number (if available)

### DIFF
--- a/snippets/csharp/programming-guide/discards/discard-pattern2.cs
+++ b/snippets/csharp/programming-guide/discards/discard-pattern2.cs
@@ -20,12 +20,12 @@ public class Example
          case IFormatProvider fmt:
             Console.WriteLine($"{fmt} object");
             break;
-         case object _:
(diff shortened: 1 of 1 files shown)


BillWagner:
I want to apologize for the lateness of my review here @psaxton I'm still thinking this through.

I like the changes you made, moving to the `switch` expressions. On the other hand, those changes don't really address the concern that this sample doesn't show a good use case for the discard pattern.

I'll keep noodling a bit.

psaxton:
This actually does though. The switch on type requires a variable to assign
the typed value to. i.e. It must be `case object _:` because `case object:`
is not syntactically valid.

Thanks,
--Phil

BillWagner:
Good point @psaxton 
I'll :shipit: now.

PR #952, opened 2019-06-05
Modified return value of SurrogateSelector example

rpetrusha:
## Modified return value of SurrogateSelector example

Fixes dotnet/dotnet-api-docs#2548

### DIFF
--- a/snippets/csharp/VS_Snippets_Remoting/SurrogateSelector/CS/SurrogateSelector.cs
+++ b/snippets/csharp/VS_Snippets_Remoting/SurrogateSelector/CS/SurrogateSelector.cs
@@ -24,7 +24,7 @@ class Employee
 sealed class EmployeeSerializationSurrogate : ISerializationSurrogate 
 {
 
-    // Serialize the Employee object to save the object’s name and address fields.
+    // Serialize the Employee object to save the objectï¿½s name and address fields.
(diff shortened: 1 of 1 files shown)


snippets/csharp/VS_Snippets_Remoting/SurrogateSelector/CS/SurrogateSelector.cs
```diff
@@ -34,7 +34,7 @@ sealed class EmployeeSerializationSurrogate : ISerializationSurrogate
         info.AddValue("address", emp.address);
     }
 
-    // Deserialize the Employee object to set the object�s name and address fields.
+    // Deserialize the Employee object to set the object�s name and address fields.
```
yasuf: I can't see the symbol for the quotes here, maybe just leave these docs lines as they were?

BillWagner:
```suggestion
    // Deserialize the Employee object to set the object's name and address fields.
```

BillWagner:
```suggestion
    // Deserialize the Employee object to set the object's name and address fields.
```

PR #955, opened 2019-06-06
Modern async

Mackiovello:
## Summary

Updated code to use modern C# constructs for async code, such as `async/await` and `Task.Run`.

Fixes https://github.com/dotnet/docs/issues/10627

### DIFF
--- a/snippets/csharp/VS_Snippets_Misc/tpl_cancellation/cs/snippet02.cs
+++ b/snippets/csharp/VS_Snippets_Misc/tpl_cancellation/cs/snippet02.cs
@@ -2,16 +2,16 @@
 using System;
 using System.Threading;
 using System.Threading.Tasks;
+
 class Program
 {
-    static void Main()
+    static async Task Main()
     {
         var tokenSource2 = new CancellationTokenSource();
         CancellationToken ct = tokenSource2.Token;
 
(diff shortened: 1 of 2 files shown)


snippets/csharp/VS_Snippets_Misc/tpl_cancellation/cs/snippet02.cs
```diff
@@ -27,19 +27,19 @@ static void Main()
...
         tokenSource2.Cancel();
 
-        // Just continue on this thread, or Wait/WaitAll with try-catch:
+        // Just continue on this thread, or await with try-catch:
         try
         {
-            task.Wait();
+            await task;
         }
         catch (AggregateException e)
```
BillWagner: By changing to `await`, the `AggregateException` isn't the right exception type. It should be an `OperationCanceledException`. 

That also means the `foreach` loop isn't needed, and a single `Console.WriteLine` is sufficient.

Mackiovello: Good point! I caught that in the other example. Now it's fixed for this one too!

BillWagner:
Thanks for making these updates @Mackiovello 

I'll :shipit: now

PR #965, opened 2019-06-07
Update links

Potapy4:
## Update links

Contributes to dotnet/docs#12641

Related to dotnet/docs#12718


Youssef1313:
This URL may also need to get updated [here](https://docs.microsoft.com/en-us/dotnet/api/system.uri?view=netframework-4.8)

rpetrusha:
It's likely that we use the contoso.com URL in far more places, @Potapy4. Rather than fix them piecemeal, it's probably best to create an issue with a list of articles/examples that use contoso.com. The issue can then be used in tracking progress as the URLs are fixed.

rpetrusha:
I'm sorry that this has languished, @Potapy4. I'll approve and merge it now.

PR #966, opened 2019-06-07
ArgumentNullException: swap string parameters

pkulikov:
When the [ArgumentNullException is constructed with two parameters](https://docs.microsoft.com/en-us/dotnet/api/system.argumentnullexception.-ctor?view=netframework-4.8#System_ArgumentNullException__ctor_System_String_System_String_), the first one is the parameter name, the second one is the exception message.

The updated snippet is used in the following docs section:
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/throw#the-throw-expression

### DIFF
--- a/snippets/csharp/language-reference/keywords/throw/coalescing.cs
+++ b/snippets/csharp/language-reference/keywords/throw/coalescing.cs
@@ -9,7 +9,7 @@ public class Person
    {
        get => name;
        set => name = value ?? 
-           throw new ArgumentNullException("Name cannot be null", nameof(value));
+           throw new ArgumentNullException(nameof(value), "Name cannot be null");
    }   
 // </Snippet1>
 }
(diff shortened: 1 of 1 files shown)


snippets/csharp/language-reference/keywords/throw/coalescing.cs
```diff
@@ -9,7 +9,7 @@ public string Name
    {
        get => name;
        set => name = value ?? 
-           throw new ArgumentNullException("Name cannot be null", nameof(value));
+           throw new ArgumentNullException(nameof(value), "Name cannot be null");
```
BillWagner:
```suggestion
           throw new ArgumentNullException(paramName: nameof(value), message: "Name cannot be null");
```

I've been using named arguments to make these examples crystal clear. Thoughts?

pkulikov: @BillWagner clearer - the better. Thanks!

PR #976, opened 2019-06-10
Added stackalloc code snippets

pkulikov:
Supports dotnet/docs#12754

New examples are to show how to allocate memory, NOT how to use it afterwards. That's why there is no output generated and they are much shorter than existing examples. Having them in the project allows to make sure that all the code compiles.

### DIFF
--- a/csharp/language-reference/operators/StackallocOperator.cs
+++ b/csharp/language-reference/operators/StackallocOperator.cs
@@ -1,0 +1,57 @@
+using System;
+
+namespace operators
+{
+    public static class StackallocOperator
+    {
+        public static void Examples()
+        {
+            Console.WriteLine("No snippets with output");
+        }
+
+        private static void AssignToPointer()
+        {
+            // <SnippetAssignToPointer>
+            unsafe
+            {
+                int length = 3;
+                int* numbers = stackalloc int[length];
+                for (var i = 0; i < length; i++)
+                {
+                    numbers[i] = i;
+                }
+            }
+            // </SnippetAssignToPointer>
+        }
+
+        private static void AssignToSpan()
+        {
+            // <SnippetAssignToSpan>
+            int length = 3;
+            Span<int> numbers = stackalloc int[length];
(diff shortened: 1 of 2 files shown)


csharp/language-reference/operators/StackallocOperator.cs
```diff
@@ -0,0 +1,57 @@
...
+                numbers[i] = i;
+            }
+            // </SnippetAssignToSpan>
+        }
+
+        private static void AsExpression()
+        {
+            // <SnippetAsExpression>
+            int length = 1000;
+            Span<byte> buffer = length <= 1024 ? stackalloc byte[length] : new byte[length];
```
BillWagner: Nice!

csharp/language-reference/operators/StackallocOperator.cs
```diff
@@ -0,0 +1,57 @@
...
+            buffer = stackalloc byte[128];
```
BillWagner: I'm not sure what the intent of this line is.  Should it be removed?

(Note: I'm looking at this before reviewing the associated doc.)

pkulikov: @BillWagner intent is to show that one can assign a new stack allocated block to an *earlier* declared variable of `Span` type. The preceding text in the article mentions assignment expression. However, I agree that snippet itself looks weak, not coherent; I was doubting as well if to include that line.

pkulikov: @BillWagner how does that line look in the context of the article? If still not fitting, I'll remove it.

BillWagner: I would remove this line from the sample:

```csharp
buffer = stackalloc byte[128];
```

The text states that assignment is legal. This extra assignment, without some use of the preceding value doesn't add anything.

pkulikov: @BillWagner I updated the snippet.

PR #997, opened 2019-06-18
Use PascalCase for function naming

Youssef1313:


Youssef1313:
> 
> 
> Thanks, @Youssef1313, for attempting to bring some consistency to the naming conventions used in this code file. There are a number of other places where camel casing instead of Pascal casing were used in public method names. It would be nice if you could change those as well. They are on lines 68, 72, 76, 97, 99, 293, 298, 302, 330, 384, 501, 696, 704, 708, 784, 804, 834, 1039, 1043, 1107, 1189, 1207, 1218, 1220, and 1230. (There are a few more that are not included in snippet tags and so don't appear in the documentation, and this list doesn't include type, property, event, or field names, which are also inconsistent.)

I've fixed the naming on those lines and some of the properties and events that I noticed.

rpetrusha:
Thank you for all of the additional corrections, @Youssef1313. We appreciate your taking the time to apply uniform naming conventions to this file. I'll merge your PR now.

PR #1023, opened 2019-06-27
typo: Obect -> Object

nschonni:


BillWagner:
@nschonni These all LGTM

However, before I merge it, is there a corresponding PR in dotnet/docs? If not, we should wait to merge this. I'm afraid that changing the snippet names would create build errors.

nschonni:
No, there are no hits in dotnet/docs. Unless it's used somewhere else, they may be orphan examples

rpetrusha:
The code can also be used in dotntet/dotnet-api-docs.

nschonni:
No hits in dotntet/dotnet-api-docs either

rpetrusha:
If not only the snippets but the source code files generally aren't used in their docs or dotnet-api-docs, I wonder whether we should delete them rather than fix typos. What do you think, @BillWagner?

BillWagner:
I was wondering the same thing @rpetrusha 

I think it would be better to remove samples that aren't used.

nschonni:
PS: I did a quick github search over on the MicrosoftDocs and Aspnet (since it owns Entity Framework), but didn't get hits either.
Feel free to close this if you want to delete the examples instead

BillWagner:
Thanks @nschonni 

I'll close this, and open a new PR to remove the affected files.

PR #1065, opened 2019-07-09
 Fix code format 

lindexi:
## Summary

Fix the code in [Assembly.GetType Method (System.Reflection)](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.gettype?view=netframework-4.8 )

In [corefx/coding-style.md at master · dotnet/corefx](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md ) NO.1, we use Allman style braces, where each brace begins on a new line. A single line statement block can go without braces but the block must be properly indented on its own line and must not be nested in other statement blocks that use braces (See issue 381 for examples). One exception is that a using statement is permitted to be nested within another using statement by starting on the following line at the same indentation level, even if the nested using contains a controlled block.


snippets/csharp/VS_Snippets_CLR_System/system.reflection.assembly.gettype/cs/gettype1.cs
```diff
@@ -4,39 +4,41 @@
...
-                                 props[ctr].PropertyType.FullName);
-               nVirtual++;
-            }
+    public static void Main()
+    {
+        Assembly assem = typeof(Example).Assembly;
+        Type t = assem.GetType("Transportation.MeansOfTransportation");
+        if (t != null)
+        {
+            Console.WriteLine("Virtual properties in type {0}:",
```
BillWagner: Consider switching to interpolated strings:

```suggestion
            Console.WriteLine($”Virtual properties in type {t.FullName}:");
```

snippets/csharp/VS_Snippets_CLR_System/system.reflection.assembly.gettype/cs/gettype1.cs
```diff
@@ -4,39 +4,41 @@
...
+                              t.FullName);
```
BillWagner: This line would need to be deleted if the preceding suggestion was accepted.

snippets/csharp/VS_Snippets_CLR_System/system.reflection.assembly.gettype/cs/gettype1.cs
```diff
@@ -4,39 +4,41 @@
...
+            PropertyInfo[] props = t.GetProperties();
+            int nVirtual = 0;
+            for (int ctr = 0; ctr < props.Length; ctr++)
+                if (props[ctr].GetMethod.IsVirtual)
+                {
+                    Console.WriteLine("   {0} (type {1})",
```
BillWagner: I’ll suggest a similar change to use interpolated strings here.

snippets/csharp/VS_Snippets_CLR_System/system.reflection.assembly.gettype/cs/gettype1.cs
lindexi: Thank you. I will switch it.

PR #1067, opened 2019-07-09
Added an example of user-defined conversions

pkulikov:
Supports dotnet/docs#13295

### DIFF
--- a/csharp/language-reference/operators/UserDefinedConversions.cs
+++ b/csharp/language-reference/operators/UserDefinedConversions.cs
@@ -1,0 +1,34 @@
+using System;
+
+public struct Digit
+{
+    private readonly byte digit;
+
+    public Digit(byte digit)
+    {
+        if (digit > 9)
+        {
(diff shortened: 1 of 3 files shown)


csharp/language-reference/operators/UserDefinedConversions.cs
```diff
@@ -0,0 +1,34 @@
+using System;
+
+public struct Digit
```
BillWagner: This could be a `readonly struct`, although that’s not what this article is discussing. 

What are your thoughts? I’m not sure I’d propose general guidance to create `readonly struct` types whenever possible, but it might be the preferred choice when a struct is designed to be immutable.

pkulikov: @BillWagner I think it would be a good addition to raise awareness about `readonly` structs. Also I've noticed that in the corefx repo (and most probably in the coreclr as well then) as many as possible structs have been marked as `readonly`. So, marking applicable struct examples in the docs as `readonly` follows the trend. I've marked both `Digit` and `Fraction` structs as `readonly`. Thanks for the notice!

PR #1073, opened 2019-07-10
Remove not used snippets

pkulikov:
PRs dotnet/docs#13295 and dotnet/docs#13247 are merged, so some code snippets have become obsolete and can be removed.


snippets/csharp/VS_Snippets_VBCSharp/csProgGuideStatements/CS/Statements.cs
```diff
@@ -80,9 +76,6 @@ public static void test()
             System.Console.WriteLine(num); // Output: 6            
             //</Snippet3>
 
-            //<Snippet4>
-            DoWork(num);
-            //</Snippet4>
```
pkulikov: While I was cleaning this file, I checked some other snippets if they were referenced. Not referenced ones are removed.

PR #1079, opened 2019-07-11
Updating snippets/[cs-vb]/System.DateTime

Youssef1313:
**Intended Changes:**
* Using string interpolation when possible.
* Following coding conventions.
* I may include some other changes for modernizing articles.

I'm planning to go for each sub directory in the `snippets` folder and do the same, but still not sure if I can do that. The `snippets` folder is huge.

### DIFF
--- a/snippets/csharp/System.DateTime/DateWithTimeZone.cs
+++ b/snippets/csharp/System.DateTime/DateWithTimeZone.cs
@@ -1,4 +1,4 @@
-﻿// <Snippet6>
+// <Snippet6>
 using System;
 
 namespace DateTimeExtensions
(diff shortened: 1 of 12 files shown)


snippets/csharp/System.DateTime/DateWithTimeZone.cs
```diff
@@ -12,10 +12,7 @@ public struct DateWithTimeZone
         public DateWithTimeZone(DateTime dateValue, TimeZoneInfo timeZone)
         {
             dt = dateValue;
-            if (timeZone == null)
-                tz = TimeZoneInfo.Local;
-            else
-                tz = timeZone;
+            tz = timeZone ?? TimeZoneInfo.Local;
```
BillWagner: Nice!

PR #1081, opened 2019-07-11
Try Fix encoding

ChrisMaddock:
The pound sign in this code sample renders incorrectly

image
https://docs.microsoft.com/en-us/dotnet/api/system.single.tostring?view=netframework-4.8

I _believe_ this is because the file was ISO-8859-1 encoded rather than UTF-8. This message popped up when trying to edit it - according to GitHub the file encoding will be automatically changed when the commit is made. 

I also had to make an unrelated change for GitHub to allow me to submit the PR.

### DIFF
--- a/snippets/csharp/VS_Snippets_CLR_System/system.Single.ToString/CS/ToString1.cs
+++ b/snippets/csharp/VS_Snippets_CLR_System/system.Single.ToString/CS/ToString1.cs
@@ -173,12 +173,12 @@ public class Class1
       // Displays:    $16,325.63
       culture = CultureInfo.CreateSpecificCulture("en-GB");
       Console.WriteLine(value.ToString(specifier, culture));
-      // Displays:    £16,325.63
+      // Displays:    Â£16,325.63
       
       specifier = "E04";
       culture = CultureInfo.CreateSpecificCulture("sv-SE");
       Console.WriteLine(value.ToString(specifier, culture));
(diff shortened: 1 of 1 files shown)


BillWagner:
Hmm. The displayed result still looks incorrect.  @rpetrusha Do you have any ideas?

sharwell:
I can take a look

rpetrusha:
The same issue, incidentally, is true of the VB example.

rpetrusha:
I'm sure that I'm missing some obvious point, @BillWagner, but in the absence of merging it and building the article where it's used, how do we know that it isn't rendering properly?

BillWagner:
@rpetrusha @ChrisMaddock When I look at the update, the £ symbol was not appearing in the first diff line. Could have been my mistake this morning. What are you seeing?

And, yes, I agree that the VB sample should also be fixed.

sharwell:
There are several issues I found. I'll submit a PR separately that you can compare with this.

rpetrusha:
I am seeing the £ symbol in the first diff line.

sharwell:
I submitted #1085 to fix encoding problems in C# and VB source files. I found 108 files impacted.

rpetrusha:
Thanks for pointing out this issue, @ChrisMaddock. Since #1085 offers a more comprehensive fix, I'll close this PR.

PR #1085, opened 2019-07-12
Fix encoding problems for C# and VB files

sharwell:
* Set repository encoding to utf-8-bom
* Save all C# and VB files as UTF-8, preserving the current byte order state

I will file a separate pull request that applies the byte order mark to all source files currently missing one.

⚠️ **DO NOT SQUASH** this pull request during the merge. The author is overridden for some commits to avoid me taking ownership of entire files.

💡 The files showing in the diff as binary files were UTF-16 encoded. After this pull request, these files will show as proper text files in the repository.


snippets/csharp/VS_Snippets_CLR/conceptual.clscompliant/cs/naming1.cs
```diff
@@ -48,7 +48,7 @@ public class Size
 //    Naming2a.cs(13,8): (Location of symbol related to previous warning)
 // </Snippet17>
 
-// Å   Å    ������
```
sharwell: ❓ What is this supposed to be? I was not able to figure out what the original intent here was.

rpetrusha: I think that it's somehow just garbage that was introduced into the file during migration. There's no snippet, it's not used in code, and I included an entry point only so that I wouldn't have to specify that I was compiling a library from the command line.

sharwell: Should I leave it like this, remove the part after the second Å, or remove the whole line?

rpetrusha: Just remove the whole line.

rpetrusha:
Thanks, @sharwell, for the final change. I'll merge your PR now.

PR #1093, opened 2019-07-17
Added the delegate operator examples

pkulikov:
Supports dotnet/docs#13432

### DIFF
--- a/csharp/language-reference/operators/DelegateOperator.cs
+++ b/csharp/language-reference/operators/DelegateOperator.cs
@@ -15,23 +15,16 @@ namespace operators
         private static void AnonymousMethod()
         {
             // <SnippetAnonymousMethod>
-            Action<IEnumerable<int>> display = delegate (IEnumerable<int> s)
-            {
-                Console.WriteLine(string.Join(" ", s));
-            };
-
-            var numbers = new[] { 1, 2, 3 };
-            display(numbers);  // output: 1 2 3
+            Func<int, int, int> sum = delegate (int a, int b) { return a + b; };
+            Console.WriteLine(sum(3, 4));  // output: 7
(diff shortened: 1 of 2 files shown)


csharp/language-reference/operators/DelegateOperator.cs
```diff
@@ -0,0 +1,53 @@
...
+
+            var numbers = new[] { 1, 2, 3 };
+            display(numbers);  // output: 1 2 3
+            // </SnippetAnonymousMethod>
+        }
+
+        private static void Lambda()
+        {
+            // <SnippetLambda>
+            Action<IEnumerable<int>> display = s => Console.WriteLine(string.Join(" ", s));
```
BillWagner: I think we need an example that declares multiple parameters, so that readers see that syntax:

```csharp
Func<int, int> sum = (a, b) => a + b;
```

I am pondering a couple other samples to add. I'll comment on those on dotnet/docs#13432

pkulikov: I'm fine with the suggestion. Still, we should remember that the article is about the delegate operator. So, main goal is to show that lambdas are more concise. The referenced article about lambdas provides more examples of them.

BillWagner: hmm. Very true. I think one example with multiple arguments is wise. That's one bit of syntax I often see questions on. The referenced article is the right location for async lambdas, lambdas with tuples, closures, and so on.

pkulikov: @BillWagner I've updated examples to use multiple arguments; those new examples make lambdas look even more better than original examples, I would say. Thanks for the note and suggestion.

PR #1104, opened 2019-07-22
Fix enum flag values

Thraka:
## Summary

Enum values were not in power of 2 but * 2 😆 

Fixes dotnet/docs#13508


Thraka:
Whoops. I confused decimal with hex. DOH!

PR #1116, opened 2019-07-26
Add examples of DateTime converters for serialization in S.Text.Json

layomia:
Adds code samples referenced in https://github.com/dotnet/docs/pull/13526.

### DIFF
--- a/snippets/standard/datetime/json/datetime-converter-examples/example1/Program.cs
+++ b/snippets/standard/datetime/json/datetime-converter-examples/example1/Program.cs
@@ -1,0 +1,78 @@
+﻿using System;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using System.Text.RegularExpressions;
+
+namespace DateTimeConverterExamples
+{
+    public class DateTimeConverterExample1 : JsonConverter<DateTime>
+    {
+        public override DateTime Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            if (reader.TokenType != JsonTokenType.String)
+            {
+                throw new JsonException();
+            }
+
+            return DateTime.Parse(reader.GetString());
+        }
+
+        public override void Write(Utf8JsonWriter writer, DateTime value, JsonSerializerOptions options)
+        {
+            writer.WriteStringValue(value.ToString());
+        }
+    }
+
+    class Program
+    {
+        private static void ParseDateTimeWithDefaultOptions_Example1()
+        {
+            var _ = JsonSerializer.Deserialize<DateTime>(@"""04-10-2008 6:30 AM"""); // Throws JsonException
+        }
+
+        private static void FormatDateTimeWithDefaultOptions()
+        {
+            Console.WriteLine(JsonSerializer.Serialize(DateTime.Parse("04-10-2008 6:30 AM -4"))); // "2008-04-10T06:30:00-04:00"
+        }
+
+        private static void ProcessDateTimeWithCustomConverter_Example1()
+        {
+            var options = new JsonSerializerOptions();
+            options.Converters.Add(new DateTimeConverterExample1());
+
+            var testDateTimeStr = "04-10-2008 6:30 AM";
+            var testDateTimeJson = @"""" + testDateTimeStr + @"""";
+
+            var resultDateTime = JsonSerializer.Deserialize<DateTime>(testDateTimeJson, options);
+            Console.WriteLine(resultDateTime); // 4/10/2008 6:30:00 AM
+
+            var resultDateTimeJson = JsonSerializer.Serialize(DateTime.Parse(testDateTimeStr), options);
+            Console.WriteLine(Regex.Unescape(resultDateTimeJson));  // "4/10/2008 6:30:00 AM"
+        }
+
+        static void Main(string[] args)
+        {
+            // Parsing non-compliant format as DateTime fails by default.
+            try
+            {
+                ParseDateTimeWithDefaultOptions_Example1();
+            }
+            catch (JsonException e)
+            {
+                Console.WriteLine(e.Message); // The JSON value could not be converted to System.DateTime. Path: $ | LineNumber: 0 | BytePositionInLine: 20.
+            }
+
+            // Formatting with default options prints according to extended ISO 8601 profile.
+            FormatDateTimeWithDefaultOptions();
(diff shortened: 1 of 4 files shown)


333fred:
@layomia did you mean to request someone else for review? I'm not familiar with this code.

snippets/standard/datetime/json/datetime-converter-examples/example1/Program.cs
```diff
@@ -0,0 +1,78 @@
+﻿using System;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using System.Text.RegularExpressions;
+
+namespace DateTimeConverterExamples
+{
+    public class DateTimeConverterExample1 : JsonConverter<DateTime>
+    {
+        public override DateTime Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
```
ahsonkhan: Should there be a validation check that `typeToConvert == typeof(DateTime)`?

In this snippet:
https://github.com/dotnet/samples/blob/ee556965f4fa50cda130816dc5ad75714bdb7b4d/snippets/standard/datetime/json/datetime-converter-examples/example1/Program.cs#L10-L18

@JeremyKuhne, @steveharter - maybe you can help answer this

snippets/standard/datetime/json/datetime-converter-examples/example2/Program.cs
```diff
@@ -0,0 +1,82 @@
...
+        public override DateTime Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            if (reader.TokenType != JsonTokenType.String)
+            {
+                throw new JsonException();
+            }
+
+            if (Utf8Parser.TryParse(reader.ValueSpan, out DateTime value, out _, 'R'))
+            {
+                reader.Read();
```
ahsonkhan: Why is there an extra `.Read()` here?

snippets/standard/datetime/json/datetime-converter-examples/example2/Program.cs
```diff
@@ -0,0 +1,82 @@
...
+    public class DateTimeConverterExample2 : JsonConverter<DateTime>
+    {
+        public override DateTime Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            if (reader.TokenType != JsonTokenType.String)
+            {
+                throw new JsonException();
+            }
+
+            if (Utf8Parser.TryParse(reader.ValueSpan, out DateTime value, out _, 'R'))
```
ahsonkhan: Can you verify if this works with data that needs to be un-escaped? I suspect it won't and we'd need `reader.GetString()` and then re-transcode to UTF-8.

If so, this highlights the need for even more. https://github.com/dotnet/corefx/issues/35386

snippets/standard/datetime/json/datetime-converter-examples/example2/Program.cs
```diff
@@ -0,0 +1,82 @@
...
+            }
+
+            throw new FormatException();
+        }
+
+        public override void Write(Utf8JsonWriter writer, DateTime value, JsonSerializerOptions options)
+        {
+            Span<byte> destination = new byte[29];
+
+            bool result = Utf8Formatter.TryFormat(value, destination, out _, new StandardFormat('R'));
```
ahsonkhan: Will this always fill the whole 29 bytes?

Maybe make it 31 and manually append the quotes at start/end (and slice on `bytesWritten + 1`)?

Then, you don't need to re-encode it as string and can write the utf-8 bytes directly:
```C#
// or something to that affect depending on how much gets written
writer.WriteStringValue(destination.Slice(0, bytesWritten + 1));
```

snippets/standard/datetime/json/datetime-converter-examples/example2/Program.cs
```diff
@@ -0,0 +1,82 @@
...
+using System.Buffers;
+using System.Buffers.Text;
+using System.Diagnostics;
+using System.Text;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace DateTimeConverterExamples
+{
+    public class DateTimeConverterExample2 : JsonConverter<DateTime>
```
ahsonkhan: nit: Instead of having names contain Example1/2, maybe make them descriptive? Up to you.

Maybe:
`DateTimeConverterCustomStandardFormat_R`

snippets/standard/datetime/json/datetime-converter-examples/example2/Program.cs
```diff
@@ -0,0 +1,82 @@
...
+        static void Main(string[] args)
+        {
+            // Parsing non-compliant format as DateTime fails by default.
+            try
+            {
+                ParseDateTimeWithDefaultOptions_Example2();
+            }
+            catch (JsonException e)
+            {
+                Console.WriteLine(e.Message); // The JSON value could not be converted to System.DateTime. Path: $ | LineNumber: 0 | BytePositionInLine: 31.
```
ahsonkhan: A lot of this code/comment ends up off to-the-side, and people have to then scroll horizontally.

nit: Consider moving it above/below the line and maybe use more locals/etc to line-wrap after a certain column width, to help with readability.

snippets/standard/datetime/json/datetime-converter-examples/example1/Program.cs
steveharter: No you don't need a check on `typeToConvert`. It will always be `typeof(DateTime)`.

The `typeToConvert` parameter is useful for polymorphic cases and when using generics to get `typeof(<T>)` value in a performant way.

snippets/standard/datetime/json/datetime-converter-examples/example2/Program.cs
layomia: No it won't - the parser expects an un-escaped input. I've refactored this.

snippets/standard/datetime/json/datetime-converter-examples/example2/Program.cs
layomia: > Will this always fill the whole 29 bytes?

It'll always be 29 per the standard format. I'll add a comment.

> Then, you don't need to re-encode it as string and can write the utf-8 bytes directly:
Nice, done!

PR #1121, opened 2019-07-29
Update tutorial code with snippets

JRAlexander:


JRAlexander:
Thanks, @luisquintanilla!

PR #1125, opened 2019-07-30
Update snippets

JRAlexander:


JRAlexander:
Yeah, debating on what goes in there... This is a new tutorial. Thanks @mairaw!

PR #1140, opened 2019-08-06
Cleanup AutoMapper usage in SimpleFeedReader

lbargaoanu:


csharp/tutorials/nullable-reference-migration/finished/SimpleFeedReader/Services/NewsService.cs
```diff
@@ -73,8 +73,6 @@ public NewsStoryProfile()
             // Create the AutoMapper mapping profile between the 2 objects.
             // ISyndicationItem.Id maps to NewsStoryViewModel.Uri.
             CreateMap<ISyndicationItem, NewsStoryViewModel>()
-                .ForCtorParam("published", opt => opt.MapFrom(src => src.Published))
```
BillWagner: @lbargaoanu Does AutoMapper manage the mapping when case is different?

lbargaoanu: Yes, AM is case insensitive.

lbargaoanu: And in fact, if there were a mismatch, `AssertConfigurationIsValid` would report it.

PR #1144, opened 2019-08-07
Replace Marshall... with Marshal...

NextTurn:
Contributes to dotnet/docs#11986

/cc @vitek-karas


NextTurn:
@rpetrusha I think there should also be a [standard](https://github.com/dotnet/samples/tree/master/snippets/standard) directory in the repo root for .NET Guide samples.

vitek-karas:
Thanks a lot @NextTurn !

rpetrusha:
> @rpetrusha I think there should also be a standard directory in the repo root for .NET Guide samples.

Yes, there should be, @NextTurn. Although there don't appear to be any samples to put in it at the moment, aside from possibly the linq samples in the core directory.

PR #1147, opened 2019-08-08
Update example.vb

ericmutta:
Convert commented-out code from C# to VB syntax.

Fixes dotnet/dotnet-api-docs#2944


ericmutta:
Thanks for the rapid response and the guidance on editing the samples, much appreciated!

PR #1152, opened 2019-08-09
Bug in method IsApproximatelyEqual fixed

wollmich:
The divisor can get negative, so we have to take as well the absolute value of the divisor and not only of the difference. Otherwise IsApproximatelyEqual(0, -1, 1e-15) would return true.


wollmich:
See as well https://github.com/dotnet/samples/pull/1153 for the Visual Basic version

rpetrusha:
Thanks for submitting #1153, which I'll review shortly. I'll merge this PR now. Thanks again.

PR #1155, opened 2019-08-12
Complete all examples in System.Text.Json's DateTime(Offset) guidance doc

layomia:
Also,
- Crucial comments that need to be localized have removed and placed in the guidance doc itself.
- Program output has been removed from the output site and is now at the bottom of each program.

### DIFF
--- a/snippets/standard/datetime/json/computing-with-jsondocument-valid.cs
+++ b/snippets/standard/datetime/json/computing-with-jsondocument-valid.cs
@@ -1,0 +1,58 @@
+﻿using System;
+using System.Text.Json;
+
+public class Example
+{
+    private static double ComputeAverageTemperatures(string json)
+    {
+        var options = new JsonDocumentOptions
+        {
+            AllowTrailingCommas = true
+        };
+
+        using (JsonDocument document = JsonDocument.Parse(json, options))
+        {
+            int sumOfAllTemperatures = 0;
+            int count = 0;
+
+            foreach (JsonElement element in document.RootElement.EnumerateArray())
+            {
+                DateTimeOffset date = element.GetProperty("date").GetDateTimeOffset();
+
+                if (date.DayOfWeek == DayOfWeek.Monday)
+                {
+                    int temp = element.GetProperty("temp").GetInt32();
+                    sumOfAllTemperatures += temp;
+                    count++;
+                }
+            }
+
+            var averageTemp = (double)sumOfAllTemperatures / count;
+            return averageTemp;
+        }
+    }
+
+    public static void Main(string[] args)
+    {
+        string json =
+                @"[" +
+                    @"{" +
+                        @"""date"": ""2013-01-07T00:00:00Z""," +
+                        @"""temp"": 23," +
+                    @"}," +
+                    @"{" +
+                        @"""date"": ""2013-01-08T00:00:00Z""," +
+                        @"""temp"": 28," +
+                    @"}," +
+                    @"{" +
+                        @"""date"": ""2013-01-14T00:00:00Z""," +
+                        @"""temp"": 8," +
+                    @"}," +
+                @"]";
+
+        Console.WriteLine(ComputeAverageTemperatures(json));
+    }
+}
+
+// The example displays the following output:
+// 15.5
--- a/snippets/standard/datetime/json/computing-with-jsondocument-valid.cs
+++ b/snippets/standard/datetime/json/computing-with-jsondocument-valid.cs
@@ -5,7 +5,7 @@ public class Example
 {
     private static double ComputeAverageTemperatures(string json)
     {
-        var options = new JsonDocumentOptions
+        JsonDocumentOptions options = new JsonDocumentOptions
         {
             AllowTrailingCommas = true
         };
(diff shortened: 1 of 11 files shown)


snippets/standard/datetime/json/computing-with-jsondocument-valid.cs
```diff
@@ -0,0 +1,58 @@
...
+                        @"""date"": ""2013-01-08T00:00:00Z""," +
+                        @"""temp"": 28," +
+                    @"}," +
+                    @"{" +
+                        @"""date"": ""2013-01-14T00:00:00Z""," +
+                        @"""temp"": 8," +
+                    @"}," +
+                @"]";
+
+        Console.WriteLine(ComputeAverageTemperatures(json));
```
ahsonkhan: It isn't super clear/obvious from looking at the two samples, why one passes and one fails. However, I would imagine the text surrounding the sample will highlight that, so its probably fine.

I understand the localization concern, but does it still make sense to put a comment within/above the json string explicitly stating that the date time format is what's difference. At a quick glance, the two samples are almost identical otherwise.

@rpetrusha, what do you think?

snippets/standard/datetime/json/reading-with-utf8jsonreader-error.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        var json = new Utf8JsonReader(utf8Data);
+        while (json.Read())
+        {
+            if (json.TokenType == JsonTokenType.String)
+            {
+                Console.WriteLine(json.TryGetDateTime(out DateTime datetime));
+                Console.WriteLine(datetime);
+
+                var _ = json.GetDateTime();
```
ahsonkhan: nit: Don't use var

snippets/standard/datetime/json/reading-with-utf8jsonreader-error.cs
```diff
@@ -0,0 +1,30 @@
...
+        }
+    }
+}
+
+// The example displays the following output:
+// False
+// 1/1/0001 12:00:00 AM
+// Unhandled exception. System.FormatException: The JSON value is not in a supported DateTime format.
+//     at System.Text.Json.Utf8JsonReader.GetDateTime()
+//     at Example.Main(String[] args) in D:\console_apps\DateTimeConverterExamples\Program.cs:line 19
```
ahsonkhan: Should we avoid callstack info like this, which is specific to a local repro (like the path info)? Is stating just the exception and the immediate method that caused it sufficient?

snippets/standard/datetime/json/reading-with-utf8jsonreader-valid.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        var json = new Utf8JsonReader(utf8Data);
+        while (json.Read())
+        {
+            if (json.TokenType == JsonTokenType.String)
+            {
+                Console.WriteLine(json.TryGetDateTime(out DateTime datetime));
+                Console.WriteLine(datetime);
+                Console.WriteLine(json.GetDateTime());
+
```
ahsonkhan: nit: remove extra line

snippets/standard/datetime/json/serializing-with-jsonserializer.cs
```diff
@@ -0,0 +1,22 @@
...
+        public string Name { get; set; }
+        public DateTime ExpiryDate { get; set; }
+    }
+
+    public static void Main(string[] args)
+    {
+        Product p = new Product();
+        p.Name = "Banana";
+        p.ExpiryDate = new DateTime(2019, 7, 26);
+        string json = JsonSerializer.Serialize(p);
```
ahsonkhan: nit: consider adding a line break between product p init and the call to serialize.

snippets/standard/datetime/json/computing-with-jsondocument-valid.cs
rpetrusha: It's fine to add a comment in source code; I probably didn't make that clear, @ahsonkhan -- I didn't mean to imply that comments shouldn't be used. What I really meant is that comments should not be the only thing that illustrates something important in the example -- it's also important that the textual explanation of the example explains why one succeeds and the other fails.

snippets/standard/datetime/json/reading-with-utf8jsonreader-error.cs
rpetrusha: We've generally displayed all of the information from Exception.Message. But a shorter version is fine, since this is really a bit excessive.

rpetrusha:
I'll approve and merge once you address the comments from @ahsonkhan, @layomia.

snippets/standard/datetime/json/reading-with-utf8jsonreader-error.cs
ahsonkhan: > We've generally displayed all of the information from Exception.Message. 

Question: And do we generally include the filepath info in such cases, like D:\console_apps\ etc which are just specific to @layomia's local run? I would have expected that to cause some confusion.

layomia:
@rpetrusha, I've pushed another commit addressing the feedback.

PR #1156, opened 2019-08-12
Updated code to compile, match description

rpetrusha:
## Updated code to compile, match description

Fixes dotnet/docs#13762


Youssef1313:
In C# sample, lines 24 & 31 are inconsistent, It will look better if both of them use the `()` or both of them doesn't use the `()`.

rpetrusha:
Thanks for the comments, @Youssef1313. I've addressed them all in the code.

PR #1160, opened 2019-08-13
Use string interpolation

Youssef1313:
- Used string interpolation
- Used var instead of actual type when it's clear from the right side
- Modernizing sample

### DIFF
--- a/snippets/visualbasic/VS_Snippets_CLR_Classic/classic Hashtable.Clear Example/VB/source.vb
+++ b/snippets/visualbasic/VS_Snippets_CLR_Classic/classic Hashtable.Clear Example/VB/source.vb
@@ -24,7 +24,7 @@ Public Class SamplesHashtable
 
         ' Displays the count and values of the Hashtable.
         Console.WriteLine("After Clear,")
-        Console.WriteLine($"   Count    : {0}")
+        Console.WriteLine($"   Count    : {myHT.Count}")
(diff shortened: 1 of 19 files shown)


Youssef1313:
@rpetrusha , I updated the samples you mentioned.
I also noticed that I get some different results when I run the code, can you check those commits ? (I still didn't check the parallel C# for those, I'll wait your review first)

snippets/visualbasic/VS_Snippets_CLR_Classic/classic Hashtable.Clear Example/VB/source.vb
```diff
@@ -25,7 +24,7 @@ Public Class SamplesHashtable
 
         ' Displays the count and values of the Hashtable.
         Console.WriteLine("After Clear,")
-        Console.WriteLine("   Count    : {0}", myHT.Count)
+        Console.WriteLine($"   Count    : {0}")
```
rpetrusha:
```suggestion
        Console.WriteLine($"   Count    : {myHT.Count}")
```

Youssef1313:
Thanks, @rpetrusha :)
I've committed it. What about the commits with message `Fixed wrong output`, should I apply same changes to the C# sample?

rpetrusha:
Thanks for the additional changes, @Youssef1313. I'll merge your PR now.

PR #1161, opened 2019-08-13
Corrected indexer set operation

rpetrusha:
## Corrected indexer set operation

Fixes dotnet/dotnet-api-docs#2952


Youssef1313:
I think these same changes can be applied here too ?
https://github.com/dotnet/samples/blob/master/snippets/csharp/VS_Snippets_CLR_System/system.Collections.Specialized.IOrderedDictionary_Implementation/cs/remarks.cs

Lines: 22, 67 ?
@rpetrusha  @BillWagner

PR #1174, opened 2019-08-16
Add dependabot config

richlander:
To enable: https://dependabot.com/dotnet/

The config is generated and must be re-generated as projects are added or directories are re-named.


richlander:
Well, it won't improve the "build process". It just "improves" the project files as a source asset. It's not conceptually different than running roslyn analyzers with fixs on (if that's a thing) and then having auto-generated PRs.

terrajobst:
Yikes. I hope they improve this experience with wildcards. Having a file with three lines per .csproj doesn't look like a great plan to me -- it feels very redundant and can easily break. 

image

mairaw:
I'm not familiar with this. What will the automated PRs contain? How does updating this file works?

richlander:
This is how it works: https://dependabot.com/#how-it-works

Yes @terrajobst, this is the pattern: https://github.com/dependabot/feedback/issues/632#issuecomment-521448392

I want to adopt this service for the dotnet org for repos where it is valuable. This is GH's service for update PackageReferences and similar things for new packages, via automated PRs.  Once we start using it, we can give feedback. Yes, this model isn't great, but it is what it is and is workable (for now), IMO. I assume they want the config.yml to avoid a lot of compute intensive tasks.

mairaw:
Cool! Thanks for sharing the additional information. I think it’s good to merge.

@Rick-Anderson you might like this for aspnet/docs.

richlander:
Merging on its own won't do anything (meaning, it's totally fine to merge w/o concern). The important step is turning on the service after that. Anyone can do that but I was going to wait for that step when I got back from vacation. I'm OOF next week. If someone wants to try it ahead of time, feel free. I was only volunteering to do that if there is a mess that needs to be cleaned up.

As an aside (and in response to @terrajobst), I might extend the tool so that it can be run in CI in a test mode (is my current config.yml still valid?). That would help mitigate some of that concern.

richlander:
I'll let @terrajobst provide any additional concerns before merging. I'm not going to merge before taking off. Feel free to do that, however.

BillWagner:
Thanks all.

I'll :shipit: now.

BillWagner:
@terrajobst Yes, I agree that wildcards would be better.

We (.NET Content team) also have a long-standing work item to continue to rationalize our samples projects. There are *way* too many very small projects.

PR #1175, opened 2019-08-16
Removed BOM from readme.md files

rpetrusha:
## Removed BOM from readme.md files

Because of the presence of the BOM, the samples were not available in Samples Browser.


docs-product:
:shipit: Ship it!

PR #1194, opened 2019-08-22
Update whitespace1.cs

noctwolf:
## Summary

Consistent with the documentation


noctwolf:
ps，vb file does not need to be fix

PR #1196, opened 2019-08-22
 Remove Imports System, part 2

Youssef1313:
Related to: https://github.com/dotnet/samples/pull/1195

Same regex is used on these directories:
- snippets/csharp
- snippets/xml
- snippets/visualbasic/api
- snippets/visualbasic/how-to
- snippets/visualbasic/System.DateTime
- snippets/visualbasic/VS_Snippets_ADO.NET


Youssef1313:
@rpetrusha , I was wondering if all of the namespaces in the following image are imported by default ?
image

so, I may try to remove all of them too.

rpetrusha:
I think that there's a difference between Visual Studio and the command line, and we've typically written Visual Basic console applications so that they can be compiled from the command line. I'd be inclined to leave in all Imports statements other than for System and Microsoft.VisualBasic.

PR #1351, opened 2019-08-27
Add samples of custom DateTime(Offset) processing with the JSON reader/writer

layomia:
These samples are used in https://github.com/dotnet/docs/pull/14056.

### DIFF
--- a/snippets/standard/datetime/json/custom-reading-with-utf8jsonreader.cs
+++ b/snippets/standard/datetime/json/custom-reading-with-utf8jsonreader.cs
@@ -1,0 +1,26 @@
+﻿using System;
+using System.Globalization;
+using System.Text;
+using System.Text.Json;
+
+public class Example
+{
+    public static void Main(string[] args)
+    {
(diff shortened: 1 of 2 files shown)


snippets/standard/datetime/json/custom-reading-with-utf8jsonreader.cs
```diff
@@ -0,0 +1,26 @@
...
+    public static void Main(string[] args)
+    {
+        byte[] utf8Data = Encoding.UTF8.GetBytes(@"""Friday, 26 July 2019 00:00:00""");
+
+        Utf8JsonReader json = new Utf8JsonReader(utf8Data);
+        while (json.Read())
+        {
+            if (json.TokenType == JsonTokenType.String)
+            {
+                string value = json.GetString();
```
ahsonkhan: I think it would be good to include what happens if you just call `json.GetDateTime()`. Documenting the exception and behavior before showing the work-around would make things clearer.

PR #1392, opened 2019-08-31
Add information for samples browser in readme

Youssef1313:


rpetrusha:
Thanks for opening this PR, @Youssef1313. 

I think that two things are necessary before we can actually recommend that code examples be integrated into the Samples Browser:
- Since the repo houses both samples and code snippets, we need to define what code might be a candidate for inclusion in the Samples Browser .
- Diagnostics tools to identify configuration issues need to be more widely available.  

For the moment, I'm going to close this PR. When some progress is made in the second area, we can reopen the PR and add criteria for Sample Browser candidates.

PR #1397, opened 2019-09-01
Modernizing code

Youssef1313:


winforms/formatting-utility/cs/Form1.cs
```diff
@@ -241,10 +238,7 @@ private void Value_TextChanged(object sender, EventArgs e)
             label.Text = String.Empty;
             valueInfo = false;
          }
-         if (String.IsNullOrEmpty(Value.Text))
-            OKButton.Enabled = false;
-         else
-            OKButton.Enabled = true;
+         OKButton.Enabled = not string.IsNullOrEmpty(Value.Text);
```
NextTurn:
```suggestion
         OKButton.Enabled = Not String.IsNullOrEmpty(Value.Text);
```

Youssef1313: Well, That's a mistake by me, but also capitalized `Not` isn't correct. it should be `!`.
I'll fix it.

PR #1411, opened 2019-09-04
Added unmanaged constructed types sample

pkulikov:
Supports dotnet/docs#14198

### DIFF
--- a/csharp/language-reference/builtin-types/UnmanagedTypes.cs
+++ b/csharp/language-reference/builtin-types/UnmanagedTypes.cs
@@ -1,3 +1,4 @@
+// <SnippetProgramExample>
 using System;
 
 public struct Coords<T>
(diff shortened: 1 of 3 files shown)


csharp/language-reference/builtin-types/UnmanagedTypes.cs
```diff
@@ -0,0 +1,24 @@
+using System;
+
+public struct Coords<T>
+{
+    public T X;
+    public T Y;
+}
+
+public class UnmanagedTypes
```
Youssef1313: I didn't test in an IDE, but I think this class can be made static.

BillWagner: It can be static, but I don't think it's necessary.

pkulikov: In other similar occasions where the full program is in snippet, it's not static. I'd follow the same pattern here.

PR #1430, opened 2019-09-06
Move P/Invokes to NativeMethods class - callback

NextTurn:

### DIFF
--- a/snippets/cpp/VS_Snippets_CLR/conceptual.interop.marshaling/cpp/callback.cpp
+++ b/snippets/cpp/VS_Snippets_CLR/conceptual.interop.marshaling/cpp/callback.cpp
@@ -6,7 +6,7 @@ using namespace System::Runtime::InteropServices;
 public delegate bool FPtr(int value);
 public delegate bool FPtr2(String^ value);
 
-public ref class LibWrap
+private ref class NativeMethods
 {
 public:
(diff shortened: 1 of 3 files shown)


snippets/cpp/VS_Snippets_CLR/conceptual.interop.marshaling/cpp/callback.cpp
```diff
@@ -6,7 +6,7 @@ using namespace System::Runtime::InteropServices;
 public delegate bool FPtr(int value);
 public delegate bool FPtr2(String^ value);
 
-public ref class LibWrap
+private ref class NativeMethods
```
BillWagner: As with #1431, I may have read this wrong, but shouldn't this be `internal` instead of `private`?

NextTurn: There is no `internal` type in C++.

For more information, see [Type visibility](https://docs.microsoft.com/cpp/dotnet/how-to-define-and-consume-classes-and-structs-cpp-cli?view=vs-2019#BKMK_Type_visibility).

BillWagner: Doh!  That explains it. I read the changes thinking it was C#. I'll review again.

PR #1431, opened 2019-09-06
Move P/Invokes to NativeMethods class - structures

NextTurn:

The C++ example seems to have adopted the old-style 80-word line length limit. Do we have a new style?


snippets/cpp/VS_Snippets_CLR/conceptual.interop.marshaling/cpp/structures.cpp
```diff
@@ -33,11 +33,11 @@ public value struct MyArrayStruct
 {
 public:
     bool flag;
-    [MarshalAs(UnmanagedType::ByValArray, SizeConst=3)]
+    [MarshalAs(UnmanagedType::ByValArray, SizeConst = 3)]
     array<int>^ vals;
 };
 
-public ref class LibWrap
+private ref class NativeMethods
```
BillWagner: I may be looking at the diffs wrong, but shouldn't this be `internal` instead of `private`?

PR #1444, opened 2019-09-09
Remove unused arg and fix indentation

imba-tjd:
Page: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/join-clause#example
Fixes https://github.com/dotnet/docs.zh-cn/issues/472

### DIFF
--- a/snippets/csharp/VS_Snippets_VBCSharp/CsCsrefQueryKeywords/CS/Join.cs
+++ b/snippets/csharp/VS_Snippets_VBCSharp/CsCsrefQueryKeywords/CS/Join.cs
@@ -26,24 +26,24 @@ namespace Joins
         // Specify the first data source.
         List<Category> categories = new List<Category>()
         { 
-            new Category(){Name="Beverages", ID=001},
-            new Category(){ Name="Condiments", ID=002},
-            new Category(){ Name="Vegetables", ID=003},
-            new Category() {  Name="Grains", ID=004},
-            new Category() {  Name="Fruit", ID=005}            
+            new Category {Name="Beverages", ID=001},
+            new Category {Name="Condiments", ID=002},
+            new Category {Name="Vegetables", ID=003},
+            new Category {Name="Grains", ID=004},
+            new Category {Name="Fruit", ID=005}            
         };
 
         // Specify the second data source.
         List<Product> products = new List<Product>()
        {
-          new Product{Name="Cola",  CategoryID=001},
-          new Product{Name="Tea",  CategoryID=001},
-          new Product{Name="Mustard", CategoryID=002},
-          new Product{Name="Pickles", CategoryID=002},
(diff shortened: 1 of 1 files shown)


snippets/csharp/VS_Snippets_VBCSharp/CsCsrefQueryKeywords/CS/Join.cs
```diff
@@ -165,7 +165,7 @@ orderby prod.CategoryID
                 Console.WriteLine("   {0}:{1}", item.ProductName, item.Category);
             }
 
-            Console.WriteLine("GroupJoin3: {0} items in 1 group", totalItems, groupJoinQuery3.Count());
+            Console.WriteLine("GroupJoin3: {0} items in 1 group", totalItems);
```
BillWagner: Can you explain why we're removing the `Count` from the output?

snippets/csharp/VS_Snippets_VBCSharp/CsCsrefQueryKeywords/CS/Join.cs
```diff
@@ -185,7 +185,7 @@ void LeftOuterJoin()
             // A nested foreach statement  is required to access group items
             foreach (var prodGrouping in leftOuterQuery)
             {
-                Console.WriteLine("Group:", prodGrouping.Count());
+                Console.WriteLine("Group:");
```
BillWagner: same question.

imba-tjd:
See the first and second output, they are correct.

```
   InnerJoin:
   Cola      1
   Tea       1
   Mustard   2
   Pickles   2
   Carrots   3
   Bok Choy  3
   Peaches   5
   Melons    5
   InnerJoin: 8 items in 1 group.


   Unshaped GroupJoin:
   Group:
       Cola      1
       Tea       1
   Group:
       Mustard   2
       Pickles   2
   Group:
       Carrots   3
       Bok Choy  3
   Group:
   Group:
       Peaches   5
       Melons    5
   Unshaped GroupJoin: 8 items in 5 unnamed groups
```

The first example has only 1 group. The second has 5 groups and they are unnamed.

Now `GroupJoin3` is indeed has only 1 group. No need to output `groupJoinQuery3.Count()`.

And `Left Outer Join` is indeed unnamed.

So the output doesn't need changing.

snippets/csharp/VS_Snippets_VBCSharp/CsCsrefQueryKeywords/CS/Join.cs
Youssef1313: @BillWagner , The old code seems to be not using the removed `Count`. There is no `{1}` in the first argument of `WriteLine`. I don't know if this is the intended behavior or not. ( I haven't read the code sample nor the article )

PR #1446, opened 2019-09-09
Fixed formatting of readme.md for DiagnosticScenarios sample

rpetrusha:
## Fixed formatting of readme.md for DiagnosticScenarios sample

Bad formatting was causing the sample to not appear in the Samples Browser.

### DIFF
--- a/core/diagnostics/DiagnosticScenarios/readme.md
+++ b/core/diagnostics/DiagnosticScenarios/readme.md
@@ -1,8 +1,8 @@
 ---
 languages:
-   - csharp
- products:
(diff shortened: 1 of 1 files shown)


Youssef1313:
That formatting came from https://github.com/dotnet/docs/issues/13921 .
The issue needs to be edited if the formatting is incorrect, so contributors who migrates some samples using the issue instructions doesn't make the same mistake.

rpetrusha:
Thanks for pointing that out, @Youssef1313. I'll fix it now.

PR #1169, opened 2019-08-14
Modernizing example

Youssef1313:
- Put curly braces on separate lines.
- Reduce nesting

PR #1474, opened 2019-09-13
Fix code error

Youssef1313:
The number between parenthesis in visual basic isn't the size of the array, it's the maximum index ( size - 1 )

### DIFF
--- a/snippets/visualbasic/VS_Snippets_Data/XmlWriter.WriteBase64/VB/writebase64.vb
+++ b/snippets/visualbasic/VS_Snippets_Data/XmlWriter.WriteBase64/VB/writebase64.vb
@@ -8,8 +8,8 @@ Class TestBase64
     Private Const bufferSize As Integer = 4096
 
     Public Shared Sub Main()
- 
-        Dim args() As String = System.Environment.GetCommandLineArgs()
+
+        Dim args As String() = System.Environment.GetCommandLineArgs()
         Dim myTestBase64 As New TestBase64()
         
         ' Check that the usage string is correct.
@@ -80,7 +80,7 @@ Class TestBase64
     Public Shared Sub DecodeOrignalObject(xmlFileName As String, fileNew As FileStream)
 
         Dim buffer(bufferSize - 1) As Byte
-        Dim readByte as integer = 0
+        Dim readByte As Integer = 0
 
         ' Create a file to write the bmp back.
         Dim bw As New BinaryWriter(fileNew)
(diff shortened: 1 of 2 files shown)


snippets/visualbasic/VS_Snippets_Data/XmlWriter.WriteBase64/VB/writebase64.vb
```diff
@@ -43,7 +43,7 @@ Class TestBase64
     ' passed by the user is encoded and included in the document.
     Public Shared Sub EncodeXmlFile(xmlFileName As String, fileOld As FileStream)
 
-        Dim buffer(bufferSize) As Byte
+        Dim buffer(bufferSize - 1) As Byte
```
BillWagner: I'll defer to @rpetrusha but I think this should not be changed. (same comment in DecodeOriginalObject.

Youssef1313: @BillWagner, In Visual Basic, We type the `size - 1` inside the parenthesis, not the `size` as in C#.

For example:

```vb
Sub Main
    Dim buffer(5) As Integer
    Console.WriteLine(buffer.Length) ' Prints: 6
End Sub

```

I actually opened issue #1481 because the array size is wrong in many visual basic code.
Hope I'm not missing something.

Youssef1313:
@rpetrusha, I haven't tested from an IDE, but why not make `CompareResult` shared and use `Module` instead of `Class` ?

Youssef1313:
I noticed that the visual basic code uses `args(1)` and `args(2)`, is it a mistake ? @rpetrusha

rpetrusha:
If you'd like to use `Module`...`End Module`, @Youssef1313, that's fine. It's even preferable.

In terms of the args array, args(0) is the executable, args(1) is the first argument, and args(2) is the second. So the code is correct.

Youssef1313:
I've used `Module` in vb and `static class` in c#.

> In terms of the args array, args(0) is the executable, args(1) is the first argument, and args(2) is the second. So the code is correct.

I haven't tested myself, but then I think the following suggested change might have been wrong ?
https://github.com/dotnet/samples/pull/1169#pullrequestreview-276637215

rpetrusha:
Thanks for making the changes, @Youssef1313. I've made one additional change -- the Shared keyword cannot be used with a method (in this case the `Main` method) for a Visual Basic module; all members are implicitly Shared. I'll merge your PR now.

PR #1478, opened 2019-09-15
Fix comments and use string interpolation

Youssef1313:


core/linq/csharp/join/LeftOuterJoin-Example-1.cs
```diff
@@ -79,15 +78,14 @@ public static void QuerySyntaxExample()
...
+                Console.WriteLine($"{v.ProductName}: {v.Category}");
             }
         }
 
-        //A left outer joinis like a cross join, except that all the left hand side elements get included at least once,
-        //even if they don't match any right hand side elements. Note how Vegetables shows up in the output even though
-        //it has no matching products.
-        //
-        //Output: 
+        // A left outer joinis like a cross join, except that all the left hand side elements get included at least once,
```
BillWagner: nit: 

```suggestion
        // A left outer join is like a cross join, except that all the left hand side elements get included at least once,
```

core/linq/csharp/join/LeftOuterJoin-Example-1.cs
```diff
@@ -6,11 +6,10 @@ namespace Join
 {
     public static class LeftOuterJoinExample1
     {
-        //A left outer joinis like a cross join, except that all the left hand side elements get included at least once,
-        //even if they don't match any right hand side elements. Note how Vegetables shows up in the output even though
-        //it has no matching products.
-        //
-        //Output: 
+        // A left outer joinis like a cross join, except that all the left hand side elements get included at least once,
```
BillWagner: nit:
```suggestion
        // A left outer join is like a cross join, except that all the left hand side elements get included at least once,
```

BillWagner:
Thanks for making these updates @Youssef1313 

I'll :shipit: now.

PR #1492, opened 2019-09-18
Fix syntax inconsistency

themindfulcoder:
Fix minor syntax inconsistencies in the snippet.

## Summary

No fundamental changes.


themindfulcoder:
@BillWagner it was awkward process to find the source from the documentation site. I had to inspect the page, copy the link to the git repo instead of being able to just use the edit button.
Not sure if there was a simpler way though.

BillWagner:
@themindfulcoder That's the best way now.  

We're working on enhancements to our build system that will provide better navigation.

ping @dend for more details.

PR #1493, opened 2019-09-18
Revise transfer learning sample, after tutorial review

natke:
- Move data classes into the main program, as it majorly simplifies the tutorial
- Use interface types rather than var where possible
- Remove references to "tuning" the TensorFlow model


machine-learning/tutorials/TransferLearningTF/Program.cs
```diff
@@ -60,53 +55,53 @@ private struct InceptionSettings
         // </SnippetInceptionSettings>
 
         // Build and train model
-        public static ITransformer ReuseAndTuneInceptionModel(MLContext mlContext, string dataLocation, string imagesFolder, string inputModelLocation, string outputModelLocation)
+        public static ITransformer GenerateModel(MLContext mlContext, string dataLocation, string imagesFolder, string inputModelLocation, string outputModelLocation)
```
luisquintanilla: Suggestion:

Call the `GenerateModel` method `TrainAndEvaluateModel` or something like that since both are happening here.

machine-learning/tutorials/TransferLearningTF/Program.cs
```diff
@@ -16,150 +13,106 @@ class Program
...
+                  labelColumnName: "LabelKey",
+                  predictedLabelColumnName: "PredictedLabel");
             // </SnippetEvaluate>
 
             //<SnippetDisplayMetrics>
             Console.WriteLine($"LogLoss is: {metrics.LogLoss}");
             Console.WriteLine($"PerClassLogLoss is: {String.Join(" , ", metrics.PerClassLogLoss.Select(c => c.ToString()))}");
             //</SnippetDisplayMetrics>
 
+            mlContext.Model.Save(model, trainingData.Schema, "model.zip");
```
luisquintanilla: Not sure if this step is going to be part of the tutorial since it does not have a snippet tag.

machine-learning/tutorials/TransferLearningTF/Program.cs
```diff
@@ -200,6 +152,37 @@ public static IEnumerable<ImageData> ReadFromTsv(string file, string folder)
...
+            [LoadColumn(0)]
+            public string ImagePath;
+
+            [LoadColumn(1)]
+            public string Label;
+        }
+        // </SnippetDeclareImageData>
+
+        // <SnippetDeclareImagePrediction>
+        public class ImagePrediction : ImageData
```
luisquintanilla: This is inheriting and works in this case because the way the pipeline is structured. Since we removed this practice from the How-To's do we still want to keep it here?

machine-learning/tutorials/TransferLearningTF/Program.cs
natke: Good catch. I added this line for debugging. Will remove it

machine-learning/tutorials/TransferLearningTF/Program.cs
natke: As discussed, we'll leave this here for now, as revise later if necessary

machine-learning/tutorials/TransferLearningTF/Program.cs
natke: You might argue that evaluation is part of training :-) 

I think I'd like to leave this as generate, as the method returns the model ...

PR #1559, opened 2019-09-24
Fix No Answer Inference

paneer:
## Summary

Use Null Conditional and Null Coalescing operators to suggest `No Answer`

Fixes dotnet/docs#14542

### DIFF
--- a/csharp/NullableIntroduction/NullableIntroduction/SurveyResponse.cs
+++ b/csharp/NullableIntroduction/NullableIntroduction/SurveyResponse.cs
@@ -14,7 +14,7 @@ namespace NullableIntroduction
 
         // <SnippetSurveyStatus>
         public bool AnsweredSurvey => surveyResponses != null;
-        public string Answer(int index) => surveyResponses.GetValueOrDefault(index, "No answer")!;
+        public string Answer(int index) => surveyResponses?.GetValueOrDefault(index) ?? "No answer";
         // </SnippetSurveyStatus>
 
         public SurveyResponse(int id) => Id = id;


csharp/NullableIntroduction/NullableIntroduction/SurveyResponse.cs
```diff
@@ -14,7 +14,7 @@ public class SurveyResponse
 
         // <SnippetSurveyStatus>
         public bool AnsweredSurvey => surveyResponses != null;
-        public string Answer(int index) => surveyResponses.GetValueOrDefault(index, "No answer")!;
+        public string Answer(int index) => surveyResponses?.GetValueOrDefault(index) ?? "No answer";
```
Thraka: I think this should be

```suggestion
        public string Answer(int index) => surveyResponses?[index] ?? "No answer";
```

paneer: @Thraka, index op on a dictionary even with a null conditional operator may result in a runtime `KeyNotFoundException`. Obviously, one could check for `ContainsKey` but the resulting code would be pretty verbose:
```suggestion
        public string Answer(int index) => surveyResponses?.GetValueOrDefault(index) ?? "No answer";
```
I simply left the extension method as it makes the code readable.

BillWagner: This is a good discussion.  @paneer is correct, the extension method is needed.

I'll mark this conversation resolved.

Thraka: Ahh yes duh on me!

PR #1571, opened 2019-09-26
Update source.cs

Youssef1313:


snippets/csharp/VS_Snippets_ADO.NET/Classic WebData DataTable.PrimaryKey Example/CS/source.cs
```diff
@@ -18,8 +18,8 @@ private void GetPrimaryKeys(DataTable table)
         columns = table.PrimaryKey;
 
         // Get the number of elements in the array.
-        Console.WriteLine("Column Count: " + columns.Length);
-        for(int i = 0; i < columns.Length; i++)
+        Console.WriteLine($"Column Count: {columns.Length}");
+        for (int i = 0; i < columns.Length; i++)
         {
             Console.WriteLine(columns[i].ColumnName + columns[i].DataType);
```
rpetrusha:
```suggestion
            Console.WriteLine($"{columns[i].ColumnName} {columns[i].DataType}");
```

Youssef1313: The original line wasn't adding a space between the two values, but I think the new suggested one does. (I think it's better to have the space)

Does that require to revise the article that uses this example to see if that space addition would make a change there ?
@rpetrusha

rpetrusha:
Thanks for making the additional changes, @Youssef1313. I'll merge your PR now.

PR #1579, opened 2019-09-27
Remove not correct comment

pkulikov:
The article in question:
https://docs.microsoft.com/en-us/dotnet/csharp/how-to/safely-cast-using-pattern-matching-is-and-as-operators

`PatternMatchingNullable` doesn't use the `as` operator as the comment says. It looks like that comment has been copied from another snippet:
https://github.com/dotnet/samples/blob/876733cac352d3497130c94a335e8eefa0812d3f/snippets/csharp/how-to/safelycast/asandis/Program.cs#L38-L42


Youssef1313:
Maybe it meant `is` ?

PR #1596, opened 2019-10-02
Fix property sample coding  conventions

rumaniel:
I saw the MSDN pages about properties and found that there is another [sample coding conventions](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/properties).
Other property page samples don't have `_` prefix but that page has.
So I fix the conventions as other samples.

### DIFF
--- a/snippets/csharp/VS_Snippets_VBCSharp/csProgGuideIndexers/CS/Indexers.cs
+++ b/snippets/csharp/VS_Snippets_VBCSharp/csProgGuideIndexers/CS/Indexers.cs
@@ -208,38 +208,38 @@ namespace Wrap2
     //<Snippet5>
     public class BaseClass
     {
-        private string name = "Name-BaseClass";
-        private string id = "ID-BaseClass";
+        private string _name = "Name-BaseClass";
+        private string _id = "ID-BaseClass";
 
         public string Name
         {
-            get { return name; }
+            get { return _name; }
             set { }
         }
 
         public string Id
         {
-            get { return id; }
+            get { return _id; }
             set { }
         }
     }
 
     public class DerivedClass : BaseClass
     {
-        private string name = "Name-DerivedClass";
-        private string id = "ID-DerivedClass";
+        private string _name = "Name-DerivedClass";
+        private string _id = "ID-DerivedClass";
 
         new public string Name
         {
             get
             {
-                return name;
+                return _name;
             }
 
-            // Using "protected" would make the set accessor not accessible. 
+            // Using "protected" would make the set accessor not accessible.
             set
             {
-                name = value;
+                _name = value;
             }
         }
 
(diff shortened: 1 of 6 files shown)


BillWagner:
Thanks for making this update @rumaniel 

Before I merge this, can you make a related PR in dotnet/docs to update the reference to the `seconds` field. The article on [properties](https://github.com/dotnet/docs/blame/master/docs/csharp/programming-guide/classes-and-structs/properties.md#L31) would need one change.

It's important that we keep the article and the sample in sync.

Youssef1313:
https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md

> We use _camelCase for internal and private fields and use readonly where possible. Prefix internal and private instance fields with _, static fields with s_ and thread static fields with t_. When used on static fields, readonly should come after static (e.g. static readonly not readonly static). Public fields should be used sparingly and should use PascalCasing with no prefix when used.

rumaniel:
Yes, there is no wrong convention. But I want to match the convention of [the overview article](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/properties) and other property articles([using](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/using-properties), [interface](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/interface-properties) and [restricting accessor](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/restricting-accessor-accessibility)).

So how about modifying other 3 articles?

BillWagner:
This is a great conversation.  I'll propose the following guidelines:

1. Most important, any variable changes should be checked with the articles where a sample is included to make sure they match after the change. Especially for readers learning the language, those details matter for the learning experience.
1. If we do make variable name changes, we should conform to the corefx coding standard.

Youssef1313:
> 
> 
> Yes, there is no wrong convention. But I want to match the convention of [the overview article](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/properties) and other property articles([using](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/using-properties), [interface](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/interface-properties) and [restricting accessor](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/restricting-accessor-accessibility)).
> 
> So how about modifying other 3 articles?

By **correct**, I really meant the convention used in corefx (Also ReSharper follows it), @rumaniel

BillWagner:
@rumaniel Are you going to create a related PR in dotnet/docs to update the article where these snippets are used? I think we've reached agreement on the naming conventions to use. We do need to ensure the article text matches the sample.

rumaniel:
Sorry for the late comments. I will update this PR and docs.

rumaniel:
I revert my origin commits and update other conventions with the corefx coding standards.

PR #1624, opened 2019-10-11
Pipe Complete() -> CompleteAsync()

halter73:
- CompleteAsync() is the newer, better API that avoids sync-over-async

See https://github.com/dotnet/docs/pull/14414#discussion_r333309579

### DIFF
--- a/snippets/csharp/pipelines/Pipe.cs
+++ b/snippets/csharp/pipelines/Pipe.cs
@@ -52,7 +52,7 @@ namespace Pipes
                 }
             }
 
-            // Tell the PipeReader that there's no more data coming.
+             // By completing PipeWriter, tell the PipeReader that there's no more data coming.
             await writer.CompleteAsync();
         }
 
--- a/snippets/csharp/pipelines/Pipe.cs
+++ b/snippets/csharp/pipelines/Pipe.cs
@@ -53,7 +53,7 @@ namespace Pipes
             }
 
             // Tell the PipeReader that there's no more data coming.
-            writer.Complete();
+            await writer.CompleteAsync();
         }
 
         async Task ReadPipeAsync(PipeReader reader)
@@ -80,7 +80,7 @@ namespace Pipes
             }
 
             // Mark the PipeReader as complete.
-            reader.Complete();
+            await reader.CompleteAsync();
         }
 
         bool TryReadLine(ref ReadOnlySequence<byte> buffer, out ReadOnlySequence<byte> line)


snippets/csharp/pipelines/Pipe.cs
```diff
@@ -53,7 +53,7 @@ async Task FillPipeAsync(Socket socket, PipeWriter writer)
             }
 
             // Tell the PipeReader that there's no more data coming.
-            writer.Complete();
+            await writer.CompleteAsync();
```
pkulikov: The comment mentions a reader, but code completes a writer. Is that correct?

davidfowl: Good catch

Rick-Anderson:
@BillWagner can we S&M this? I need to PR the code to clean up the formatting to get rid of horizontal scroll bars on tablets.

snippets/csharp/pipelines/Pipe.cs
halter73: I think the comment is fine as is. The call to CompleteAsync is "telling" the other side, the PipeReader, "that there's no more data coming."

pkulikov: > I think the comment is fine as is. The call to CompleteAsync is "telling" the other side, the PipeReader, "that there's no more data coming."

I've thought that's the case. Is that fine to make it more explicit:
```
// By completing PipeWriter, we tell the PipeReader that there's no more data coming.
```

halter73: Sounds good to me. Open a PR.

halter73: Sounds good to me. You should open a PR.

snippets/csharp/pipelines/Pipe.cs
```diff
@@ -53,7 +53,7 @@ async Task FillPipeAsync(Socket socket, PipeWriter writer)
             }
 
             // Tell the PipeReader that there's no more data coming.
```
pkulikov: > Sounds good to me. You should open a PR.

@halter73 would the suggestion within this PR work?

```suggestion
             // By completing PipeWriter, we tell the PipeReader that there's no more data coming.
```

BillWagner:
I accepted @pkulikov 's suggestion on the comment.

I'll :shipit: now.

/cc @Rick-Anderson

snippets/csharp/pipelines/Pipe.cs
halter73: Thanks!

PR #1627, opened 2019-10-11
Example for "SortedSet<T>" does not compile.

doterik:
'var' may only appear within a local variable:
```
public class ByFileExtension : IComparer<string>
{
    ...
    var caseiComp = new CaseInsensitiveComparer();

```
Use 'CaseInsensitiveComparer', or move into:
```
public int Compare(string x, string y)
{
    ...
```

While I was on it, also double rows of ```s.EndsWith(".doc")```.

### DIFF
--- a/snippets/csharp/VS_Snippets_CLR_System/system.collections.generic.sortedset/cs/program.cs
+++ b/snippets/csharp/VS_Snippets_CLR_System/system.collections.generic.sortedset/cs/program.cs
@@ -111,7 +111,6 @@ class Program
     {
         s = s.ToLower();
         return (s.EndsWith(".txt") ||
-            s.EndsWith(".doc") ||
             s.EndsWith(".xls") ||
             s.EndsWith(".xlsx") ||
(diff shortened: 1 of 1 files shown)


doterik:
I would like to extend this a bit further.

My language is not native English so maybe this is not an issue but when I look at the example for ```GetViewBetween(T, T)``` I become unsure of the upper and lower bounds, are they both inclusive or not, especially when I think of the new Range in C# 8.

In the document for _GetViewBetween_, I got the impression they both are inclusive, as in SQL and for me common speech but the example states that ```GetViewBetween("avi", "avj")``` returns a view of all AVI files. Well, that may be true, but what if I have AVJ files, quite unusual but never the less, wouldn't I receive these also, or?

I think this could be more clearly explained.

BillWagner:
@doterik 

> I would like to extend this a bit further.

I like all your suggestions in this comment. Would you like to open a new PR with these changes? If so, I'll watch for it.

PR #1676, opened 2019-10-27
Modernize code sample

Youssef1313:
I don't see a good reason to have a class called `StudentClass` and a class called `Student` nested in it.
I've combined them in a `Student` class.

I've also renamed `ID` to `Id` to follow [Capitalization Conventions](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions#capitalizing-compound-words-and-common-terms).

### DIFF
--- a/snippets/csharp/concepts/linq/how-to-query-a-collection-of-objects_1.cs
+++ b/snippets/csharp/concepts/linq/how-to-query-a-collection-of-objects_1.cs
@@ -57,7 +57,7 @@
             return avg > 0 ? (int)avg / 10 : 0;
         }
 
-        public void QueryHighScores(int exam, int score)
+        public static void QueryHighScores(int exam, int score)
         {
             var highScores = from student in students
                              where student.ExamScores[exam] > score
@@ -74,8 +74,7 @@
     {
         public static void Main()
         {
-            var sc = new Student();
-            sc.QueryHighScores(1, 90);
+            Student.QueryHighScores(1, 90);
 
             // Keep the console window open in debug mode.
             Console.WriteLine("Press any key to exit");
--- a/snippets/csharp/concepts/linq/how-to-query-a-collection-of-objects_1.cs
+++ b/snippets/csharp/concepts/linq/how-to-query-a-collection-of-objects_1.cs
@@ -1,65 +1,61 @@
-﻿    public class StudentClass
+﻿    public class Student
     {
         #region data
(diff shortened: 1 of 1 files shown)


snippets/csharp/concepts/linq/how-to-query-a-collection-of-objects_1.cs
```diff
@@ -78,7 +74,7 @@ public class Program
     {
         public static void Main()
         {
-            StudentClass sc = new StudentClass();
+            var sc = new Student();
```
BillWagner: Consistent with the change I requested above, the `sc` object isn't needed, and the next line would be `Student.QueryHighScores(1,90);`

snippets/csharp/concepts/linq/how-to-query-a-collection-of-objects_1.cs
```diff
@@ -1,66 +1,62 @@
...
+        // Helper method, used in GroupByRange.
         protected static int GetPercentile(Student s)
         {
             double avg = s.ExamScores.Average();
             return avg > 0 ? (int)avg / 10 : 0;
         }
 
-        
-
         public void QueryHighScores(int exam, int score)
```
BillWagner: After these changes, this class is more consistent if the `QueryHighScores` method is `static`.

Youssef1313:
```suggestion
        public static void QueryHighScores(int exam, int score)
```

snippets/csharp/concepts/linq/how-to-query-a-collection-of-objects_1.cs
Youssef1313:
```suggestion
```

snippets/csharp/concepts/linq/how-to-query-a-collection-of-objects_1.cs
```diff
@@ -78,7 +74,7 @@ public class Program
     {
         public static void Main()
         {
-            StudentClass sc = new StudentClass();
+            var sc = new Student();
             sc.QueryHighScores(1, 90);
```
Youssef1313:
```suggestion
            Student.QueryHighScores(1, 90);
```

BillWagner:
Thanks for making the additional updates. I'll :shipit: now.

PR #1682, opened 2019-10-31
Update Interest Rate Sample

SeanKilleen:
## Summary

* [x] Change `InterestRate` --> `InterestRatePercent` so that we're not trying to deal with conversion in a way that muddies the sample
* [x] Update the values to `7.5` and `7.1` respectively, rather than `.075` and `.071`

At this point the sample code should be consistent with what is outputted in the docs.

Fixes https://github.com/dotnet/docs/issues/14367

### DIFF
--- a/csharp/serialization/Loan.cs
+++ b/csharp/serialization/Loan.cs
@@ -12,7 +12,7 @@ namespace serialization
    public class Loan : INotifyPropertyChanged
     {
         public double LoanAmount { get; set; }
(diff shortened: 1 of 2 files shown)


Esjihn:
Thanks for the hard work you two @BillWagner and @SeanKilleen. Its awesome to see suggestions from the community get considered and implemented. 

Best Regards, 
Matthew D. Miller
Information Systems Programmer
Florida Senate

PR #1686, opened 2019-11-01
Add contextual understanding for logging delegates example

SeanKilleen:
## Note: this will be in conjunction with a fix in the docs
We'll be creating a new file here and changing some pointers, so this PR will be linked to another PR in that repo and both should be merged at the same time.

Link to the doc fix PR: https://github.com/dotnet/docs/pull/15612

## The Issue
The logging delegates example uses `+= LogToConsole`, but the documentation does not provide context as to where that static method exists (it's housed in `Program.cs` in the example, but that's not evident in the docs or the region that's selected for display.)

## Proposed Solution
Creating a separate static class, e.g `LoggingMethods`, to house this example should mean that the docs can be updated to explicitly show the contents of the class and where it's being used, as well as a clearer conceptual understanding of what's going on.

Fixes https://github.com/dotnet/docs/issues/14103


SeanKilleen:
These two PRs are ready for review alongside each other.

PR #1687, opened 2019-11-01
Add empty Form1 load method to avoid auto-generated code causing an error

SeanKilleen:
## Summary

When users click on the form in the example associated with this code, code will be generated that looks for a load method. Because one is not defined, the sample will then error.

This PR adds that empty method, along with an explanatory comment.

Fixes https://github.com/dotnet/docs/issues/13317


SeanKilleen:
@Thraka whoops! Right you are. :)

Moved it to the location you specified. 👍

PR #1690, opened 2019-11-01
Update PrimeService_IsPrimeShould.cs

Youssef1313:
Supports https://github.com/dotnet/docs/pull/15628

### DIFF
--- a/core/getting-started/unit-testing-using-nunit/PrimeService.Tests/PrimeService_IsPrimeShould.cs
+++ b/core/getting-started/unit-testing-using-nunit/PrimeService.Tests/PrimeService_IsPrimeShould.cs
@@ -1,3 +1,4 @@
+#region Sample_FirstTest
 using NUnit.Framework;
 using Prime.Services;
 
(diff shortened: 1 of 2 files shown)


core/getting-started/unit-testing-using-nunit/PrimeService/PrimeService.cs
```diff
@@ -5,12 +5,12 @@ namespace Prime.Services
     public class PrimeService
     {
         public bool IsPrime(int candidate)
-	    {
+        {
```
Youssef1313: nit: Convert a tab to 4 spaces.

core/getting-started/unit-testing-using-nunit/PrimeService.Tests/PrimeService_IsPrimeShould.cs
```diff
@@ -6,9 +7,10 @@ namespace Prime.UnitTests.Services
     [TestFixture]
     public class PrimeService_IsPrimeShould
     {
-        private readonly PrimeService _primeService;
+        private PrimeService _primeService;
```
BillWagner: Why are you removing the `readonly` modifier here?

Youssef1313: Because `_primeService` gets instantiated in the `SetUp` method.

PR #1696, opened 2019-11-04
Fixed parameter to be ref

ericjohannsen:
## Summary

POSIX defines the parameters in DisplayEntry and the callback as `StatClass stat *`, so C# must declare them as `ref`.

Fixes #42250


Thraka:
@ericjohannsen @sharwell 
This has fallen off the radar. Is this still valid?

We just copied the snippets area into the conceptual repo dotnet/docs

PR #1739, opened 2019-11-08
Add missing F# example for various Console methods

jpierson:
I found that the F# documentation was defaulting to a C++ example so I tried my hand at doing for the most part what was a wrote translation of the C# example into F#. I tested the results and made sure to clean up a few warnings due to the nature of F# language differences. 

There is still more room for improvement to make the example more idiomatic F# but as far as that goes so does the C# example which looks like it was largely a translation for a C++ sample.

## Summary

Add missing F# example for various Console methods.


jpierson:
Interesting, curious what call is not supported. I did notice that both the C# and F# version. Only seemed to work properly when run directly in cmd and not the VSCode terminal or Powershell so maybe it's specific to some of the Console methods.

PR #1781, opened 2019-11-20
Simplify code with use conditional invoke

SuperNova911:
Contributes to [dotnet/docs#15768](https://github.com/dotnet/docs/issues/15758)

Use conditional invoke to simplify code

### DIFF
--- a/snippets/csharp/VS_Snippets_VBCSharp/csrefKeywordsModifiers/CS/csrefKeywordsModifiers.cs
+++ b/snippets/csharp/VS_Snippets_VBCSharp/csrefKeywordsModifiers/CS/csrefKeywordsModifiers.cs
@@ -227,7 +227,7 @@ namespace csrefKeywordsModifiers
         // to enable derived classes to raise the event.
         protected virtual void RaiseSampleEvent()
         {
-            // Raise the event by using the () operator.
+            // Raise the event in a thread-safe manner using the ?. operator.
             SampleEvent?.Invoke(this, new SampleEventArgs("Hello"));
         }
     }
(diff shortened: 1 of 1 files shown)


snippets/csharp/VS_Snippets_VBCSharp/csrefKeywordsModifiers/CS/csrefKeywordsModifiers.cs
```diff
@@ -228,8 +228,7 @@ public class Publisher
         protected virtual void RaiseSampleEvent()
         {
             // Raise the event by using the () operator.
```
BillWagner: I like this change. We should change the comment to match the new code: 

```suggestion
            // Raise the event in a thread-safe manner using the ?. operator.
```

SuperNova911: You're right. How did I miss that XD

BillWagner:
Thanks @SuperNova911 I'll :shipit: now.

snippets/csharp/VS_Snippets_VBCSharp/csrefKeywordsModifiers/CS/csrefKeywordsModifiers.cs
```diff
@@ -227,7 +227,7 @@ public class Publisher
         // to enable derived classes to raise the event.
         protected virtual void RaiseSampleEvent()
         {
-            // Raise the event by using the () operator.
+            // Raise the event in a thread-safe manner using the ?. operator.
```
idbrii: How is this threadsafe? Isn't it just null-safe?

PR #1790, opened 2019-11-23
Remove extra-space from readme

Youssef1313:
This sample didn't appear in samples browser. Hopefully, removing this extra space will fix that.

I've deleted files from async-and-await, I believe they're related to the mef simple calculator sample and they already exist there.


async/async-and-await/vb/readme.md
```diff
@@ -1,15 +1,15 @@
...
+products:
 - dotnet-core
 - windows
 page_type: sample
 name: "Async Sample: Asynchronous Programming with Async and Await"
 urlFragment: "async-and-await-vb"
 description: "A .NET Core WPF application that contains the example method from Asynchronous Progamming with Async and Await in Visual Basic."
 ---
-# Introduction
+# Async Sample: Asynchronous Programming with Async and Await in Visual Basic
```
Youssef1313: I'm not sure why this heading was changed in #1773. This heading is at the very top and it will look strange to be "Introduction"

Compare both:

image

***

image


***

If you would like to include an "Introduction" heading. It will be better to be an H2.

PR #1792, opened 2019-11-25
Include closing curly brace in snippet

tdykstra:


snippets/core/system-text-json/csharp/JsonDocumentWriteJson.cs
```diff
@@ -47,8 +47,8 @@ public static void Run()
                 writer.WriteEndObject();
 
                 writer.Flush();
-                // </SnippetSerialize>
             }
+            // </SnippetSerialize>
```
BillWagner: This is fine, so I'll approve it as is.

Another alternative is to remove the brackets and change the `using` statements to the C# 8 syntax:

```csharp
using FileStream fs = File.Create(outputFileName);
using var writer = new Utf8JsonWriter(fs, options: writerOptions);
using JsonDocument document = JsonDocument.Parse(jsonString, documentOptions);
```

PR #1793, opened 2019-11-25
Fix bug in sample code

tdykstra:
The ValueSpan.SequenceEquals API doesn't convert escaped JSON values to their text equivalents, so ValueTextEquals is a better choice.

### DIFF
--- a/snippets/core/system-text-json/csharp/Utf8ReaderFromFile.cs
+++ b/snippets/core/system-text-json/csharp/Utf8ReaderFromFile.cs
@@ -8,8 +8,6 @@ namespace SystemTextJsonSamples
     class Utf8ReaderFromFile
     {
         private static readonly byte[] s_nameUtf8 = Encoding.UTF8.GetBytes("name");
-        private static readonly byte[] s_universityUtf8 = Encoding.UTF8.GetBytes("University");
-
         public static void Run()
         {
             // Read as UTF-16 and transcode to UTF-8 to convert to a Span<byte>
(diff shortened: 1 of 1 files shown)


ahsonkhan:
> I didn't know that either. I'm glad you made that change.

Yep, we added that to the remarks to help raise visibility. Hopefully that helps (and I am open to other suggestions):
https://docs.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonreader.valuespan?view=netcore-3.1#remarks
> Since this property returns the raw bytes, avoid using it for text comparison. Instead call ValueTextEquals which unescapes the text if necessary.

snippets/core/system-text-json/csharp/Utf8ReaderFromFile.cs
```diff
@@ -8,8 +8,6 @@ namespace SystemTextJsonSamples
     class Utf8ReaderFromFile
     {
         private static readonly byte[] s_nameUtf8 = Encoding.UTF8.GetBytes("name");
-        private static readonly byte[] s_universityUtf8 = Encoding.UTF8.GetBytes("University");
-
         public static void Run()
         {
             // Read as UTF-16 and transcode to UTF-8 to convert to a Span<byte>
```
ahsonkhan:
```suggestion
             // Read as UTF-16 and transcode to UTF-8 to convert to a ReadOnlySpan<byte>
```

snippets/core/system-text-json/csharp/Utf8ReaderFromFile.cs
```diff
@@ -36,11 +34,11 @@ public static void Run()
                         total++;
```
ahsonkhan: Unrelated to this change, but the following ctor is more usable/approachable, given the scenario:

```diff
- var json = new Utf8JsonReader(jsonReadOnlySpan, isFinalBlock: true, state: default);
+ var json = new Utf8JsonReader(jsonReadOnlySpan);
```

ahsonkhan: Also, it might be worth adding a comment for the `ReadAllBytes` sample, that make sure the file doesn't contain the UTF-8 BOM bytes, and if they do, slice them off before passing the text to the `Utf8JsonReader` since it expects text. Otherwise, it throws since that is considered invalid JSON.

PR #1802, opened 2019-11-27
Fixed #1801 : Can't install in VS 2019

### LINKED ISSUE #1801: Roslyn-SDK example vsixes do not install in VS 2019
The Roslyn SDK VSIX versions are out of date, and prevent installing in current versions of VS 2019. We need to update the various manifests to allow installation in more recent versions. FYI @jmarolf.

J-kit:
## Summary

Fixes #1801

### DIFF
--- a/csharp/roslyn-sdk/Tutorials/MakeConst/MakeConst.Vsix/source.extension.vsixmanifest
+++ b/csharp/roslyn-sdk/Tutorials/MakeConst/MakeConst.Vsix/source.extension.vsixmanifest
@@ -1,22 +1,22 @@
-<?xml version="1.0" encoding="utf-8"?>
-<PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
-  <Metadata>
-    <Identity Id="MakeConst.35f903bc-07ae-4a28-97ce-9235da0c513c" Version="1.0" Language="en-US" Publisher="wiwagn"/>
-    <DisplayName>MakeConst</DisplayName>
-    <Description xml:space="preserve">This is a sample diagnostic extension for the .NET Compiler Platform ("Roslyn").</Description>
-  </Metadata>
-  <Installation>
-        <InstallationTarget Id="Microsoft.VisualStudio.Community" Version="[16.0, 17.0)" />
-        <InstallationTarget Version="[16.0,17.0)" Id="Microsoft.VisualStudio.Enterprise" />
-  </Installation>
-  <Dependencies>
-    <Dependency Id="Microsoft.Framework.NDP" DisplayName="Microsoft .NET Framework" d:Source="Manual" Version="[4.5,)" />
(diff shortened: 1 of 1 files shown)


csharp/roslyn-sdk/Tutorials/MakeConst/MakeConst.Vsix/source.extension.vsixmanifest
```diff
@@ -6,7 +6,8 @@
     <Description xml:space="preserve">This is a sample diagnostic extension for the .NET Compiler Platform ("Roslyn").</Description>
   </Metadata>
   <Installation>
-    <InstallationTarget Id="Microsoft.VisualStudio.Community" Version="[15.0,)" />
+        <InstallationTarget Id="Microsoft.VisualStudio.Community" Version="[16.0, 17.0)" />
```
333fred: Indentation is off, and we should probably leave this one at `[15.0,)`. iirc that should be an open-ended range and not block 16.0.

csharp/roslyn-sdk/Tutorials/MakeConst/MakeConst.Vsix/source.extension.vsixmanifest
```diff
@@ -15,8 +16,7 @@
     <Asset Type="Microsoft.VisualStudio.MefComponent" d:Source="Project" d:ProjectName="MakeConst" Path="|MakeConst|"/>
     <Asset Type="Microsoft.VisualStudio.Analyzer" d:Source="Project" d:ProjectName="MakeConst" Path="|MakeConst|"/>
   </Assets>
-  <Prerequisites>
-    <Prerequisite Id="Microsoft.VisualStudio.Component.CoreEditor" Version="[15.0,16.0)" DisplayName="Visual Studio core editor" />
-    <Prerequisite Id="Microsoft.VisualStudio.Component.Roslyn.LanguageServices" Version="[15.0,16.0)" DisplayName="Roslyn Language Services" />
-  </Prerequisites>
+    <Prerequisites>
```
333fred: Indentation is also off here. Please restore the `Roslyn.LanguageServices` component, and make them both `[15.0,17.0)`. Perhaps @jmarolf can confirm whether we can use an open-ended range here as well to prevent this from happening again in the future.

jmarolf: The only change that is needed here is for the version number to be extended to include `17.0`

J-kit: Yea, somehow just removing the version ranges never worked for me, maybe bad luck ¯\_(ツ)_/¯

Now it works for me aswell
Can i leave it open ranged or is the 17.0 necessary?

jmarolf: I think leaving in open is fine here. In 100 years if VS `200.0` isn't compatible we can come back and update the documentation

PR #1826, opened 2019-12-07
Fixed bug in getting the image from the resx

Thraka:
## Summary

These threw an exception specifying that the resource was not a string but an image and needed to be used with `GetObject`. I've changed the code.


mjrousos:
This unusual code was needed previously because deserializing objects from resources directly didn't work on .NET Core (dotnet/corefx#26745). Since that's no longer the case, we should update or remove the comment on the method you changed since it no longer applies. 

Depending on how much code would need updated, we could even consider removing this method entirely and just calling `GetObject` directly, instead.

Thraka:
Sounds good.

PR #1837, opened 2019-12-11
Samples for Newtonsoft migration doc

tdykstra:


Thraka:
I didn't download and run the code. I didn't see any glaring errors.

tdykstra:
@Thraka Thanks for the review!

snippets/core/system-text-json/csharp/DeserializeNullToNonnullableType.cs
```diff
@@ -0,0 +1,47 @@
...
+            string jsonString;
+            var wf = WeatherForecastFactories.CreateWeatherForecast();
+
+            var serializeOptions = new JsonSerializerOptions();
+            serializeOptions.WriteIndented = true;
+            serializeOptions.Converters.Add(new DateTimeOffsetNullHandlingConverter());
+            jsonString = JsonSerializer.Serialize(wf, serializeOptions);
+            Console.WriteLine($"JSON with valid Date:\n{jsonString}\n");
+
+            var deserializeOptions = new JsonSerializerOptions();
```
ahsonkhan: I don't think it is necessary to create both serialization and deserialization specific options.

You can have one `options` in the sample and re-use for both sides. It doesn't look like the sample requires two.

snippets/core/system-text-json/csharp/DeserializeNullToNonnullableType.cs
```diff
@@ -0,0 +1,47 @@
...
+            deserializeOptions.Converters.Add(new DateTimeOffsetNullHandlingConverter());
+            wf = JsonSerializer.Deserialize<WeatherForecast>(jsonString, deserializeOptions);
+            wf.DisplayPropertyValues();
+
+            jsonString = @"{""Date"": null,""TemperatureCelsius"": 25,""Summary"":""Hot""}";
+            Console.WriteLine($"JSON with null Date:\n{jsonString}\n");
+
+            // The missing-date JSON deserializes with error if the converter isn't used.
```
ahsonkhan: The other option would be to change your POCO to have a nullable struct that can handle null (i.e. `DateTimeOffset?`).

snippets/core/system-text-json/csharp/DeserializeNullToNonnullableType.cs
```diff
@@ -0,0 +1,47 @@
...
+            try
+            {
+                wf = JsonSerializer.Deserialize<WeatherForecast>(jsonString);
+            }
+            catch (Exception ex)
```
ahsonkhan: I wouldn't catch the generic `Exception`. Can you catch the more specific exception? I think `JsonException`.

snippets/core/system-text-json/csharp/DeserializeNullToNonnullableType.cs
```diff
@@ -0,0 +1,47 @@
...
+            {
+                Console.WriteLine($"Exception thrown: {ex.Message}\n");
+            }
+
+            Console.WriteLine("Deserialize with converter");
+            deserializeOptions = new JsonSerializerOptions();
+            deserializeOptions.Converters.Add(new DateTimeOffsetNullHandlingConverter());
```
ahsonkhan: Same issue here. Is re-creating the options necessary? Re-use existing options.

snippets/core/system-text-json/csharp/DeserializeNullToNonnullableType.cs
ahsonkhan: Comment should say `null` date. Not `missing-date`. This scenario is different from the required property scenario below, where the `date` is actually missing (and not null).

snippets/core/system-text-json/csharp/DeserializeRequiredProperty.cs
```diff
@@ -0,0 +1,50 @@
...
+            wf.DisplayPropertyValues();
+
+            Console.WriteLine("Deserialize with converter");
+            try
+            {
+                deserializeOptions = new JsonSerializerOptions();
+                deserializeOptions.Converters.Add(new WeatherForecastRequiredPropertyConverter());
+                wf = JsonSerializer.Deserialize<WeatherForecast>(jsonString, deserializeOptions);
+            }
+            catch (Exception ex)
```
ahsonkhan: I would catch the most specific exception possible in all these samples.

snippets/core/system-text-json/csharp/DeserializeNullToNonnullableType.cs
ahsonkhan: Same applies elsewhere. Only have different options if they need to be specific.

snippets/core/system-text-json/csharp/ImmutablePoint.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace SystemTextJsonSamples
+{
+    // <SnippetImmutablePoint>
```
ahsonkhan: What is the purpose of such comments?

snippets/core/system-text-json/csharp/ImmutablePoint.cs
```diff
@@ -0,0 +1,19 @@
...
+        private readonly int _x;
+        private readonly int _y;
+
+        public ImmutablePoint(int x, int y)
+        {
+            _x = x;
+            _y = y;
+        }
+
+        public int X { get { return _x; } }
```
ahsonkhan: nit:
```suggestion
        public int X => _x;
```

snippets/core/system-text-json/csharp/RoundtripImmutableStruct.cs
```diff
@@ -0,0 +1,35 @@
...
+        public static void Run()
+        {
+            string jsonString;
+            var point1 = new ImmutablePoint(1, 2);
+            var point2 = new ImmutablePoint(3, 4);
+            var points = new List<ImmutablePoint> { point1, point2 };
+
+            var serializeOptions = new JsonSerializerOptions();
+            serializeOptions.WriteIndented = true;
+            //serializeOptions.Converters.Add(new WeatherForecastRequiredPropertyConverter());
```
ahsonkhan: remove commented out code

snippets/core/system-text-json/csharp/ImmutablePointConverter.cs
```diff
@@ -0,0 +1,112 @@
...
+                throw new JsonException();
+            }
+        }
+
+        public override void Write(
+            Utf8JsonWriter writer,
+            ImmutablePoint value,
+            JsonSerializerOptions options)
+        {
+            // Don't pass in options when recursively calling Serialize.
```
ahsonkhan: Should we state why? cc @steveharter

snippets/core/system-text-json/csharp/ImmutablePointConverter.cs
```diff
@@ -0,0 +1,112 @@
+﻿using System;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace SystemTextJsonSamples
+{
+    public class ImmutablePointConverter : JsonConverter<ImmutablePoint>
```
ahsonkhan: I don't know if this is the right place for this, but this converter can be written more efficiently. So, maybe we suggest that?

1) `reader.GetString()` for property name matching isn't necessary. You could use `ValueTextEquals`
2) Comparison with `JsonEncodedText` is faster than comparison with `string`.
3) Rather than asking for an int converter every time, it's probably better to get it once and either pass it in, or store it as a field by getting it in the custom ctor. cc @steveharter, @layomia 
4) You can remove one of the checks at the end (`if (!xSet || !ySet)`) by crafting the conditions differently.

```C#
public class ImmutablePointNewConverter : JsonConverter<ImmutablePoint>
{
    private readonly JsonEncodedText XName = JsonEncodedText.Encode("X");
    private readonly JsonEncodedText YName = JsonEncodedText.Encode("Y");

    private readonly JsonConverter<int> _intConverter;

    public ImmutablePointNewConverter(JsonSerializerOptions options)
    {
        if (options?.GetConverter(typeof(int)) is JsonConverter<int> intConverter)
        {
            _intConverter = intConverter;
        }
        else
        {
            throw new InvalidOperationException();
        }
    }

    public override ImmutablePoint Read(
        ref Utf8JsonReader reader,
        Type typeToConvert,
        JsonSerializerOptions options)
    {
        if (reader.TokenType != JsonTokenType.StartObject)
        {
            throw new JsonException();
        };

        int x = default;
        bool xSet = false;

        int y = default;
        bool ySet = false;

        // Get the first property.
        reader.Read();
        if (reader.TokenType != JsonTokenType.PropertyName)
        {
            throw new JsonException();
        }

        if (reader.ValueTextEquals(XName.EncodedUtf8Bytes))
        {
            x = ReadProperty(ref reader, options);
            xSet = true;
        }
        else if (reader.ValueTextEquals(YName.EncodedUtf8Bytes))
        {
            y = ReadProperty(ref reader, options);
            ySet = true;
        }
        else
        {
            throw new JsonException();
        }

        // Get the second property.
        reader.Read();
        if (reader.TokenType != JsonTokenType.PropertyName)
        {
            throw new JsonException();
        }

        if (xSet && reader.ValueTextEquals(YName.EncodedUtf8Bytes))
        {
            y = ReadProperty(ref reader, options);
        }
        else if (ySet && reader.ValueTextEquals(XName.EncodedUtf8Bytes))
        {
            x = ReadProperty(ref reader, options);
        }
        else
        {
            throw new JsonException();
        }

        reader.Read();

        if (reader.TokenType != JsonTokenType.EndObject)
        {
            throw new JsonException();
        }

        return new ImmutablePoint(x, y);
    }

    private int ReadProperty(ref Utf8JsonReader reader, JsonSerializerOptions options)
    {
        Debug.Assert(reader.TokenType == JsonTokenType.PropertyName);

        reader.Read();
        return _intConverter.Read(ref reader, typeof(int), options);
    }

    public override void Write(
        Utf8JsonWriter writer,
        ImmutablePoint value,
        JsonSerializerOptions options)
    {
        // Don't pass in options when recursively calling Serialize.
        JsonSerializer.Serialize(writer, value);
    }
}
```

snippets/core/system-text-json/csharp/ImmutablePointConverter.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            if (reader.TokenType != JsonTokenType.EndObject)
+            {
+                throw new JsonException();
+            }
+
+            return new ImmutablePoint(x, y);
+        }
+
+        public int ReadProperty(ref Utf8JsonReader reader, JsonSerializerOptions options)
```
ahsonkhan: Don't make this public.
```suggestion
        private int ReadProperty(ref Utf8JsonReader reader, JsonSerializerOptions options)
```

snippets/core/system-text-json/csharp/ImmutablePointConverter.cs
```diff
@@ -0,0 +1,112 @@
...
+            }
+            else if (propertyName == YName)
+            {
+                y = ReadProperty(ref reader, options);
+                ySet = true;
+            }
+            else
+            {
+                throw new JsonException();
+            }
```
ahsonkhan: If we are keeping this, move it to a private helper method to avoid duplication.

snippets/core/system-text-json/csharp/RoundtripCallbacks.cs
```diff
@@ -0,0 +1,36 @@
...
+
+            // <SnippetSerialize>
+            var serializeOptions = new JsonSerializerOptions();
+            serializeOptions.Converters.Add(new WeatherForecastCallbacksConverter());
+            serializeOptions.WriteIndented = true;
+            jsonString = JsonSerializer.Serialize(wf, serializeOptions);
+            // </SnippetSerialize>
+
+            Console.WriteLine($"JSON output:\n{jsonString}\n");
+            //jsonString = @"{""Date"": null,""TemperatureCelsius"": 25,""Summary"":""Hot""}";
```
ahsonkhan: Remove commented out code

snippets/core/system-text-json/csharp/RoundtripImmutableStruct.cs
```diff
@@ -0,0 +1,35 @@
...
+            deserializeOptions.Converters.Add(new ImmutablePointConverter());
+            points = JsonSerializer.Deserialize<List<ImmutablePoint>>(jsonString, deserializeOptions);
+            Console.WriteLine("Deserialized object values");
+            foreach (ImmutablePoint point in points)
+            {
+                Console.WriteLine($"X,Y = {point.X},{point.Y}");
+            }
+        }
+    }
+
```
ahsonkhan: nit: remove extra new line

snippets/core/system-text-json/csharp/RoundtripLongToString.cs
```diff
@@ -0,0 +1,25 @@
...
+namespace SystemTextJsonSamples
+{
+    public class RoundtripLongToString
+    {
+        public static void Run()
+        {
+            string jsonString;
+
+            // Serialize to create input JSON
+            var weatherForecast = WeatherForecastFactories.CreateWeatherForecastWithLong();
```
ahsonkhan: nit: If we were to follow dotnet/runtime guidelines, don't use `var` here since it isn't clear from R.H.S. what the type actually is.

snippets/core/system-text-json/csharp/RoundtripLongToString.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using System.Text.Json;
+
+namespace SystemTextJsonSamples
+{
+    public class RoundtripLongToString
+    {
+        public static void Run()
+        {
+            string jsonString;
```
ahsonkhan: nit: Move the variable declaration down to line 18?

snippets/core/system-text-json/csharp/RoundtripToFileAsync.cs
```diff
@@ -29,6 +30,24 @@ public static async Task RunAsync()
             }
             // </SnippetDeserialize>
             weatherForecast.DisplayPropertyValues();
+
+            using (FileStream fs = File.Create(fileNameUtf8))
+            {
+                using (StreamWriter writer = new StreamWriter(fs, Encoding.UTF8))
```
ahsonkhan: What's the point of this `writer`?

snippets/core/system-text-json/csharp/ImmutablePointConverter.cs
layomia: Do we have guidance against this? We do it internally: https://github.com/dotnet/runtime/blob/f77914d4c9045a01b3a91b63c28805f24850c274/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/Converters/JsonValueConverterKeyValuePair.cs#L127

snippets/core/system-text-json/csharp/RoundtripToFileAsync.cs
```diff
@@ -29,6 +30,24 @@ public static async Task RunAsync()
...
+                {
+                    await JsonSerializer.SerializeAsync(fs, weatherForecast);
+                }
+            }
+            Console.WriteLine($"The result is in {fileNameUtf8}\n");
+            using (FileStream fs = File.OpenRead(fileNameUtf8))
+            {
+                //System.Text.Json.JsonException: '0xEF' is invalid after a single JSON value. Expected end of data.
+                //weatherForecast = await JsonSerializer.DeserializeAsync<WeatherForecast>(fs);
```
ahsonkhan: Why is this commented out? Are you getting a `JsonException`?

snippets/core/system-text-json/csharp/WeatherForecastRequiredPropertyConverter.cs
```diff
@@ -0,0 +1,33 @@
...
+            JsonSerializerOptions options)
+        {
+            // Don't pass in options when recursively calling Deserialize.
+            WeatherForecast value = JsonSerializer.Deserialize<WeatherForecast>(ref reader);
+
+            // Check for required fields set by values in JSON
+            if (value.Date == default)
+            {
+                throw new JsonException("Required property not received in the JSON");
+            };
```
ahsonkhan: nit: Is this semi-colon needed?

snippets/core/system-text-json/csharp/WeatherForecastRuntimeIgnoreConverter.cs
```diff
@@ -0,0 +1,70 @@
...
+            while (reader.Read())
+            {
+                if (reader.TokenType == JsonTokenType.EndObject)
+                {
+                    return wf;
+                }
+
+                if (reader.TokenType == JsonTokenType.PropertyName)
+                {
+                    var propertyName = reader.GetString();
```
ahsonkhan: nit: don't use var here.
```suggestion
                    string propertyName = reader.GetString();
```

snippets/core/system-text-json/csharp/WeatherForecastRuntimeIgnoreConverter.cs
```diff
@@ -0,0 +1,70 @@
...
+                }
+            }
+
+            throw new JsonException();
+        }
+
+        public override void Write(Utf8JsonWriter writer, WeatherForecast wf, JsonSerializerOptions options)
+        {
+            // Location for OnSerializing "callback" code.
+            Console.WriteLine("OnSerializing");
```
ahsonkhan: Do we need this here for this sample?

snippets/core/system-text-json/csharp/ImmutablePointConverter.cs
ahsonkhan: Well, it would cause a stackoverflow in this case. I don't know if this is a bug or expected.

https://github.com/dotnet/runtime/blob/f77914d4c9045a01b3a91b63c28805f24850c274/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/JsonPropertyInfoNotNullable.cs#L89
>	System.Text.Json.Tests.dll!System.Text.Json.Serialization.Tests.ReadValueTests.ImmutablePointConverter.Write(System.Text.Json.Utf8JsonWriter writer, System.Text.Json.Serialization.Tests.ReadValueTests.ImmutablePoint value, System.Text.Json.JsonSerializerOptions options) Line 1611	C#
 	System.Text.Json.dll!System.Text.Json.JsonPropertyInfoNotNullable<object, System.Text.Json.Serialization.Tests.ReadValueTests.ImmutablePoint, System.Text.Json.Serialization.Tests.ReadValueTests.ImmutablePoint>.OnWrite(ref System.Text.Json.WriteStackFrame current, System.Text.Json.Utf8JsonWriter writer) Line 91	C#
 	System.Text.Json.dll!System.Text.Json.JsonPropertyInfo.Write(ref System.Text.Json.WriteStack state, System.Text.Json.Utf8JsonWriter writer) Line 571	C#
 	System.Text.Json.dll!System.Text.Json.JsonSerializer.Write(System.Text.Json.Utf8JsonWriter writer, int originalWriterDepth, int flushThreshold, System.Text.Json.JsonSerializerOptions options, ref System.Text.Json.WriteStack state) Line 37	C#
 	System.Text.Json.dll!System.Text.Json.JsonSerializer.WriteCore(System.Text.Json.Utf8JsonWriter writer, object value, System.Type type, System.Text.Json.JsonSerializerOptions options) Line 139	C#
 	System.Text.Json.dll!System.Text.Json.JsonSerializer.WriteValueCore(System.Text.Json.Utf8JsonWriter writer, object value, System.Type type, System.Text.Json.JsonSerializerOptions options) Line 107	C#
 	System.Text.Json.dll!System.Text.Json.JsonSerializer.Serialize<System.Text.Json.Serialization.Tests.ReadValueTests.ImmutablePoint>(System.Text.Json.Utf8JsonWriter writer, System.Text.Json.Serialization.Tests.ReadValueTests.ImmutablePoint value, System.Text.Json.JsonSerializerOptions options) Line 22	C#

In the key value pair case, looks like we only call it for `object` and end up calling `GetType` on it the next time, so it doesn't usually stackoverflow.

But, this sample breaks that:
```C#
var kvp = new KeyValuePair<object, int>("foo", 0);
KeyValuePair<object, int> parent = default;

for (int i = 0; i < 1_000; i++)
{
    parent = new KeyValuePair<object, int>(kvp, i);
    kvp = parent;
}

Console.WriteLine(JsonSerializer.Serialize(parent));
```

snippets/core/system-text-json/csharp/DeserializeNullToNonnullableType.cs
tdykstra: In most cases they have to be different variables because they are separate snippets in the doc.  If I used the same options variable I wouldn't be able to declare the variable in the second snippet., which appears as an independent snippet in the doc. The alternative would be to have separate snippets to declare the options variable and use it, i.e., 3 instead of 2 snippets n the code file and 4 instead of 2 in the doc. I can do that but splitting what is really one snippet into two separate ones would look strange in the doc.  In the instances where that's not an issue I'm changing the code to use a single options variable.

snippets/core/system-text-json/csharp/DeserializeNullToNonnullableType.cs
tdykstra: >The other option would be to change your POCO to have a nullable struct that can handle null (i.e. DateTimeOffset?).

I mention that in the text of the doc.

snippets/core/system-text-json/csharp/ImmutablePoint.cs
tdykstra: This is how we mark sections of code for showing in the doc.  Each snippet is defined by putting a start tag like this before it and an end tag like `// </SnippetNameOfSnippet>` after it.  If a code file doesn't have any of those tags, either the entire file is shown in the doc as a  single snippet or it's not used in the doc at all.

snippets/core/system-text-json/csharp/RoundtripToFileAsync.cs
tdykstra: This code isn't displayed in the doc. My intent was to provide a snippet showing how to write a file encoded as UTF-8 and to get this code validated or corrected before showing it in the doc.

snippets/core/system-text-json/csharp/RoundtripToFileAsync.cs
tdykstra: Yes, I was getting an exception when trying to read a file encoded as UTF-8 with the preceding code.

snippets/core/system-text-json/csharp/ImmutablePointConverter.cs
tdykstra: I explain the stack overflow problem in the article text.  Steve explains it in [this comment](https://github.com/dotnet/corefx/issues/36639#issuecomment-501433729) in issue 36639

tdykstra:
@ahsonkhan @steveharter @layomia Updates in response to feedback are in https://github.com/dotnet/samples/pull/1871

snippets/core/system-text-json/csharp/ImmutablePointConverter.cs
tdykstra: The corrected code is here:  https://github.com/dotnet/samples/blob/5dee66d52080f302775790bea4a25fe5afa938ac/snippets/core/system-text-json/csharp/ImmutablePointConverter.cs

snippets/core/system-text-json/csharp/RoundtripToFileAsync.cs
ahsonkhan: Are you sure? That would be a bug. I just tried it and the UTF-8 BOM bytes are correctly getting ignored. Can you share the repro file and sample where you are seeing this issue?

snippets/core/system-text-json/csharp/RoundtripToFileAsync.cs
ahsonkhan: You can write data directly to the `FileStream` using the JsonSerializer/Utf8JsonWriter APIs. That's how you emit UTF-8 encoded data. You don't need `StreamWriter`.

snippets/core/system-text-json/csharp/RoundtripToFileAsync.cs
tdykstra: These samples are in a project that runs through all of the samples when you run it. If you have this branch locally you can open the project and uncomment the `weatherForeast=await...` line, run the project, and you'll see the error.

snippets/core/system-text-json/csharp/RoundtripToFileAsync.cs
tdykstra: When I change the code to use JsonSerializer directly, the exception goes away. It only happens with the file created by the StreamWriter specifying UTF-8 encoding.  I'm fixing related parts of the doc now.

ahsonkhan: > the exception goes away. It only happens with the file created by the StreamWriter specifying UTF-8 encoding

Which exception?

snippets/core/system-text-json/csharp/RoundtripToFileAsync.cs
tdykstra: I'm pushing a commit that will eliminate this code that throws the exception - if you haven't tried it and still want to, get the code at this commit -- https://github.com/tdykstra/dotnet-samples/tree/b1c945b488b5966f816e92ddac7c002f9656b635/snippets/core/system-text-json/csharp

ahsonkhan: When I try to run it, I get the following errors:
```text
E:\GitHub\Fork\samples\snippets\core\system-text-json\csharp>dotnet run

RoundtripDictionaryTkeyEnumTValue.cs(6,18): error CS0101: The namespace 'SystemTextJsonSamples' already contains a definition for 'RoundtripDictionaryTkeyEnumTValue' [E:\GitHub\Fork\samples\snippets\core\system-text-json\csharp\SystemTextJsonSamples.csproj]
DeserializeInferredTypesToObject.cs(6,18): error CS0101: The namespace 'SystemTextJsonSamples' already contains a definition for 'DeserializeInferredTypesToObject' [E:\GitHub\Fork\samples\snippets\core\system-text-json\csharp\SystemTextJsonSamples.csproj]
SerializePolymorphic.cs(6,11): error CS0101: The namespace 'SystemTextJsonSamples' already contains a definition for 'SerializePolymorphic' [E:\GitHub\Fork\samples\snippets\core\system-text-json\csharp\SystemTextJsonSamples.csproj]
RoundtripDictionaryTkeyEnumTValue.cs(8,28): error CS0111: Type 'RoundtripDictionaryTkeyEnumTValue' already defines a member called 'Run' with the same parameter types [E:\GitHub\Fork\samples\snippets\core\system-text-json\csharp\SystemTextJsonSamples.csproj]
DeserializeInferredTypesToObject.cs(8,28): error CS0111: Type 'DeserializeInferredTypesToObject' already defines a member called 'Run' with the same parameter types [E:\GitHub\Fork\samples\snippets\core\system-text-json\csharp\SystemTextJsonSamples.csproj]
SerializePolymorphic.cs(8,28): error CS0111: Type 'SerializePolymorphic' already defines a member called 'Run' with the same parameter types [E:\GitHub\Fork\samples\snippets\core\system-text-json\csharp\SystemTextJsonSamples.csproj]
```

tdykstra: samples master has a temporary change in it that was needed to fix a docs build bug but broke the project, this PR fixes it and the docs build.  Sorry for the confusion.  I should have mentioned that  you needed specifically the PR branch, which is in my fork.

ahsonkhan: OK, I found out the issue. When you use `Encoding.UTF8` within a `StreamWriter`, it says that it is going to emit a UTF-8 BOM.
https://source.dot.net/#System.Private.CoreLib/StreamWriter.cs,81
https://source.dot.net/#System.Private.CoreLib/UTF8Encoding.cs,62

This wouldn't be a problem if the UTF-8 BOM was written at the beginning. The `JsonSerializer` ignores it, correctly.

However, because you are **not** using the `StreamWriter` and instead writing to the `FileStream` directly, the JSON payload gets written **first**, and the BOM bytes get written after, when the `StreamWriter` is disposed. The `JsonSerializer.Deserializer` then correctly throws `JsonException` since you have invalid JSON now.

Can't have byte order mark at the end of the payload (hence invalid).
image

Also note, you likely want the filename to contain the `.json` extension.
https://github.com/dotnet/samples/blob/bc25d342d4cf041b953665cb5235e8d5489d34a0/snippets/core/system-text-json/csharp/RoundtripToFileAsync.cs#L14

Removing the `StreamWriter` should and does fix the underlying issue :)

tdykstra: And in case you missed the note I sent just as I received yours, the code that generates the exception is gone from the PR branch but can be found here: https://github.com/tdykstra/dotnet-samples/tree/b1c945b488b5966f816e92ddac7c002f9656b635/snippets/core/system-text-json/csharp
But given that there's no need to wrap a StreamWriter the way I was doing it, it may not be worth investigating further.

tdykstra: Emails crossed again, got the resolution notice, thanks !

ahsonkhan: Yep, sounds good. I understood what you were getting at. Looks like there is no underlying product issue and the sample has been fixed.

snippets/core/system-text-json/csharp/WeatherForecastRequiredPropertyConverter.cs
```diff
@@ -0,0 +1,33 @@
...
+        public override WeatherForecast Read(
+            ref Utf8JsonReader reader,
+            Type type,
+            JsonSerializerOptions options)
+        {
+            // Don't pass in options when recursively calling Deserialize.
+            WeatherForecast value = JsonSerializer.Deserialize<WeatherForecast>(ref reader);
+
+            // Check for required fields set by values in JSON
+            if (value.Date == default)
```
ahsonkhan: I think this could be problematic. If the `Date` did exist in the JSON payload, but just happen to be the default value, then we still throw an exception.

For example, if `jsonString` was the following (or you serialize a POCO containing  `DateTime date = default`).
```json
{
  "Date": "0001-01-01T00:00:00+00:00",
  "TemperatureCelsius": 25,
  "Summary": "Hot"
}
```

Not sure if there is a great way to do this without using `nullable`, at least for value types like `DateTime`, int, etc.

snippets/core/system-text-json/csharp/WeatherForecastRequiredPropertyConverter.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    public class WeatherForecastRequiredPropertyConverter : JsonConverter<WeatherForecast>
+    {
+        public override WeatherForecast Read(
+            ref Utf8JsonReader reader,
+            Type type,
+            JsonSerializerOptions options)
+        {
+            // Don't pass in options when recursively calling Deserialize.
+            WeatherForecast value = JsonSerializer.Deserialize<WeatherForecast>(ref reader);
```
ahsonkhan: Re-iterating https://github.com/dotnet/docs/pull/16225#discussion_r362137385 since it makes more sense in the samples PR:

> Looks like using `value` as parameter names isn't great in the docs. They show up colored blue which makes C# harder to read. Maybe use `result` or `forecast` (or maybe this issue can be fixed)?
> image

snippets/core/system-text-json/csharp/WeatherForecastRequiredPropertyConverter.cs
tdykstra: Do you think it better to add some text warning about the case where the date has the default value in the JSON, or  remove  the converter as a workaround from the doc?

ahsonkhan: I can't think of a great/general purpose workaround for `Required` properties atm. **For now, text warning about the case is probably fine.** I will have to think about the scenario some more. One option could be to create a sentinel default value and check against that (like for an int property that is known to be positive, choose -1), but again, you wouldn't know if the payload actually contained that value or it just didn't exist.

Even for reference types (or nullable), how do you differentiate between the property not existing vs one that exists but is explicitly set to null.

ahsonkhan: @steveharter, @layomia - do you have any ideas or suggestions?

PR #1871, opened 2019-12-25
Updates to System.Text.Json samples

tdykstra:
Related to https://github.com/dotnet/docs/pull/16225


snippets/core/system-text-json/csharp/ImmutablePointConverter.cs
```diff
@@ -27,42 +41,35 @@ public class ImmutablePointConverter : JsonConverter<ImmutablePoint>
...
+            if (reader.TokenType != JsonTokenType.EndObject)
             {
                 throw new JsonException();
             }
 
-            // Get the second property.
+            return new ImmutablePoint(x, y);
+        }
+
+        private void ReadXorY(ref Utf8JsonReader reader, JsonSerializerOptions options, ref int x, ref bool xSet, ref int y, ref bool ySet)
```
ahsonkhan: I believe this is incorrect.

The following would succeed, incorrectly: `{"X": 1, "X": 1}`

The second call to `ReadXorY` needs different conditions (from https://github.com/dotnet/samples/pull/1837#discussion_r361062959). Note the `xSet` and `ySet` being used.
```C#
        if (xSet && reader.ValueTextEquals(YName.EncodedUtf8Bytes))
        {
            y = ReadProperty(ref reader, options);
        }
        else if (ySet && reader.ValueTextEquals(XName.EncodedUtf8Bytes))
        {
            x = ReadProperty(ref reader, options);
        }
        else
        {
            throw new JsonException();
        }
```

In this case, I would remove the `ReadXorY` method and just inline the logic (like you had before). Passing the bools by ref also makes it hard to reason about. The two "if blocks" are not exactly identical in my sample, so it doesn't help to move it into a helper.

snippets/core/system-text-json/csharp/ImmutablePointConverter.cs
```diff
@@ -71,42 +78,31 @@ public class ImmutablePointConverter : JsonConverter<ImmutablePoint>
...
-            {
-                throw new JsonException();
-            }
-
-            return new ImmutablePoint(x, y);
+            return _intConverter.Read(ref reader, typeof(int), options);
         }
 
-        public int ReadProperty(ref Utf8JsonReader reader, JsonSerializerOptions options)
+        private void WriteProperty(Utf8JsonWriter writer, int value, JsonEncodedText name, JsonSerializerOptions options)
```
ahsonkhan: nit: invert the order of parameters. `name` first then `value`.

snippets/core/system-text-json/csharp/WeatherForecastCallbacksConverter.cs
```diff
@@ -11,14 +11,13 @@ public class WeatherForecastCallbacksConverter : JsonConverter<WeatherForecast>
             Type type,
             JsonSerializerOptions options)
         {
-            // Place "before" code here (OnDeserializing), but note that there is no access here to the POCO instance.
-            Console.WriteLine("OnDeserializing");
```
ahsonkhan: Why were the `Console.WriteLines` removed?

ahsonkhan: My comment here (https://github.com/dotnet/samples/pull/1837#discussion_r361065269) was specifically about `WeatherForecastRuntimeIgnoreConverter.cs` which didn't need these. It makes sense to have these in the `callback` sample.

snippets/core/system-text-json/csharp/RoundtripToFileAsync.cs
```diff
@@ -30,24 +29,6 @@ public static async Task RunAsync()
             }
             // </SnippetDeserialize>
             weatherForecast.DisplayPropertyValues();
-
-            using (FileStream fs = File.Create(fileNameUtf8))
```
ahsonkhan: I see what you were trying to get at here. If the file is UTF-8 encoded (with/without BOM), the `JsonSerializer` APIs will work just fine (that is what you are doing above - that's the default behavior/expectation).

If, however, the file/stream contained data that was encoded in something else (say UTF-16), then, the `JsonSerializer APIs will fail and throw (both when the UTF-16 BOM exists or not). 

```text
// If UTF-16 BOM exists, it is considered invalid
Unhandled exception. System.Text.Json.JsonException: '0xFF' is an invalid start of a value. Path: $ | LineNumber: 0 | BytePositionInLine: 0.

// Without a UTF-16 BOM, 0x00 is invalid JSON since the JSON APIs assume UTF-8 bytes.
Unhandled exception. System.Text.Json.JsonException: '0x00' is an invalid start of a property name. Expected a '"'. Path: $ | LineNumber: 0 | BytePositionInLine: 1.
```

For reading such files, you have to read them as strings first, and you can't use the streaming APIs for that. Similarly, for writing such files, you have to first write to a UTF-8 byte[]/string first, and then write that to the file.

**I don't know if such a code sample is necessary since I don't think it will be all that common.**

snippets/core/system-text-json/csharp/RoundtripImmutableStruct.cs
```diff
@@ -31,5 +28,4 @@ public static void Run()
             }
         }
     }
-
```
ahsonkhan: nit: I only mentioned it once, but other files have extra new lines too.

For example: https://github.com/dotnet/samples/blob/bc25d342d4cf041b953665cb5235e8d5489d34a0/snippets/core/system-text-json/csharp/SerializeRuntimePropertyExclusion.cs#L36

Consider fixing it for all the files added in https://github.com/dotnet/samples/pull/1837

snippets/core/system-text-json/csharp/WeatherForecastCallbacksConverter.cs
tdykstra: Got it, will restore them.

snippets/core/system-text-json/csharp/ConvertPolymorphic.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace SystemTextJsonSamples
 {
-    class SerializePolymorphic
+    public class ConvertPolymorphic
```
ahsonkhan: nit: Other classes need `public` as well.

Like `SerializeCustomEncoding`, `RoundtripCamelCasePropertyNames`, `JsonDocumentWriteJson`, etc. Fix all?

snippets/core/system-text-json/csharp/ImmutablePoint.cs
```diff
@@ -3,17 +3,14 @@
     // <SnippetImmutablePoint>
     public struct ImmutablePoint
```
BillWagner: I'd consider making this a readonly struct:

```suggestion
    public readonly struct ImmutablePoint
```

snippets/core/system-text-json/csharp/PersonConverterWithTypeDiscriminator.cs
```diff
@@ -40,16 +40,16 @@ public override Person Read(ref Utf8JsonReader reader, Type typeToConvert, JsonS
                 throw new JsonException();
             }
 
-            Person value;
+            Person person;
             TypeDiscriminator typeDiscriminator = (TypeDiscriminator)reader.GetInt32();
             switch (typeDiscriminator)
```
BillWagner: This would be a great use for the pattern matching switch expression:

```csharp
Person person = typeDiscriminator switch
{
    TypeDiscriminator.Customer => new Customer();
    TypeDiscriminator.Employee => new Employee();
    _ => throw new JsonException();
};
```

tdykstra: Done, except I had to comma-separate the lines instead of terminating them with semicolon. 😄 

BTW, this pattern doesn't seem to be in the [C# guide switch article](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/switch) , but it is in the [C# 8 what's new](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8).

ahsonkhan:
I didn't realize there was a json file checked-in here. Why did it show up as binary?
image

In any case, should we point to the reference of where we got this snippet from (either in the comment within the JSON or as a note in the text before/after it)?

https://github.com/dotnet/samples/blob/9ce7c1bb8dd1850c2e56fe24359a9444a5d2d4b3/snippets/core/system-text-json/csharp/Universities.json#L1

https://github.com/dotnet/samples/blob/9ce7c1bb8dd1850c2e56fe24359a9444a5d2d4b3/core/json/Program.cs#L18-L19

> The JSON data used for the samples was borrowed from https://github.com/Hipo/university-domains-list under the MIT License (MIT).

tdykstra:
It's UTF-16 LE, and it only has fictional data.

ahsonkhan:
>  it only has fictional data.

Ah gotcha. Didn't realize that.

PR #1877, opened 2020-01-03
Fix interpolated string

Youssef1313:


snippets/csharp/language-reference/keywords/switch/type-pattern.cs
```diff
@@ -10,15 +10,15 @@ static void Main(string[] args)
     {
         int[] values = { 2, 4, 6, 8, 10 };
         ShowCollectionInformation(values);
-        
+
         var names = new List<string>();
         names.AddRange( new string[] { "Adam", "Abigail", "Bertrand", "Bridgette" } );
```
Youssef1313:
```suggestion
        names.AddRange(new string[] { "Adam", "Abigail", "Bertrand", "Bridgette" });
```

snippets/csharp/language-reference/keywords/switch/type-pattern.cs
```diff
@@ -28,28 +28,26 @@ private static void ShowCollectionInformation(object coll)
...
-               break;   
+               break;
             case IList list:
                Console.WriteLine($"{list.Count} items");
                break;
             case IEnumerable ie:
                string result = "";
                foreach (var e in ie) 
-                  result += "${e} ";
+                  result += $"{e} ";
```
Youssef1313: This is the fix intended in the PR (in this file and the other two files), other changes are just whitespace changes.

PR #1879, opened 2020-01-03
Make missing parallel change to vb sample

Youssef1313:
PR #1878 missed applying the same change in vb parallel sample. I've also fixed indentation.


TheRealMaxion:
Ah yes, my bad on missing that, thanks for catching it.

PR #1885, opened 2020-01-05
Delete solution files that contain only one project

Youssef1313:


BillWagner:
This LGTM @Youssef1313 

I'd like @natke to look and make sure these aren't needed for the ML.NET audience.

luisquintanilla:
Looks good to me. Do the ML.NET samples run without the solution?

Youssef1313:
@luisquintanilla, I didn't try to compile them locally, but I think they should work.

Problems with deleting solution files should **only occur** when the solution file is handling multiple projects.

Issue #1884, opened 2020-01-05
Make samples/snippets compilable/analyzable


NextTurn:
I'm about to submitting a series of PRs to check syntax and styles of samples. This issue is for general suggestions and discussions.

NextTurn:
I'm not making parallel changes to C++/VB samples... If anyone is interested, feel free to help synchronize the changes.

PR #1888, opened 2020-01-05
[C#] Split snippets into files

NextTurn:
Contributes to #1884


Youssef1313:
It seems that `Process_StandardOutput` folder isn't referenced. So, it might just get deleted ?

NextTurn:
> It seems that `Process_StandardOutput` folder isn't referenced. So, it might just get deleted ?

Fine. You probably have deleted that folder in another PR, so I'll just revert the changes.

Youssef1313:
> Fine. You probably have deleted that folder in another PR, so I'll just revert the changes.

My current open PR still didn't address this folder. But will in a later part of the deletion process. Reverting the changes you made or deleting the folder, both options seem okay to me.

PR #1895, opened 2020-01-07
Update .editorconfig

Youssef1313:
Fixes dotnet/docs#16302

I copied the exact same file from runtime repo, with small change adding vbproj.

I've also kept the following part, which isn't not part of the runtime editorconfig file.

```ini
[*.{cs,vb}]
charset = utf-8-bom
```

---

Above is for the first commit only. Later commits has more slight changes


paul1956:
In VS 2016 Preview 16.5.0 Preview 1.0 the option below is unknown
```
dotnet_code_quality_unused_parameters = non_public:suggestion
```
The line below gives 2 warning about the space after the, at a random location in the file. Removing the space and both warning go away and stay away even after adding the space back.
```
[*.{cmd, bat}]
```

PR #1897, opened 2020-01-08
Update sample code for new Migrate from Newtonsoft article

tdykstra:
To be merged before https://github.com/dotnet/docs/pull/16225


snippets/core/system-text-json/csharp/IMyInterface.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace SystemTextJsonSamples
+{
+    public interface IMyInterface
```
ahsonkhan: This sample has "placeholder" names for the types. It would be better to have a more scenario driven name (similar to WeatherForecast) that the reader can more easily understand. I can try to come up with something if you like, but using some canonical sample example that gets used to explain folks about interfaces would work better.

snippets/core/system-text-json/csharp/IForecast.cs
```diff
@@ -0,0 +1,25 @@
...
+
+    public class Forecast : IForecast
+    {
+        public DateTimeOffset Date { get; set; }
+        public int TemperatureCelsius { get; set; }
+        public string Summary { get; set; }
+        public int WindSpeed { get; set; }
+    }
+
+    public class Forecasts
```
ahsonkhan: nice!

snippets/core/system-text-json/csharp/SerializePolymorphic.cs
```diff
@@ -14,35 +14,80 @@ public static void Run()
...
+            {
+                WriteIndented = true
+            };
+            Console.WriteLine(JsonSerializer.Serialize(forecasts, options));
+            // </SnippetSerializeInterface>
+
+            WeatherForecastWithPrevious weatherForecastWithPrevious = WeatherForecastFactories.CreateWeatherForecastWithPrevious();
+            WeatherForecastWithPreviousAsObject weatherForecastWithPreviousAsObject = WeatherForecastFactories.CreateWeatherForecastWithPreviousAsObject();
+
+            Console.WriteLine();
```
ahsonkhan: Do we have such new lines elsewhere? It doesn't look like it (at least on this file). Why is it needed here? Does the output look better because of it (and if so, it makes sense to do so, but then why not elsewhere)?

tdykstra: Elsewhere I kept the WriteLine calls out of the snippets and added `\n` where needed to make the output readable.  In this instance I left the WriteLine in the snippet but didn't want to append a newline to it that would be visible in the doc so I added the call to WriteLine. For consistency with the other snippets I could send the serialization output in this snippet to jsonString instead of WriteLine.

ahsonkhan: Ah yes, you mean the previous output on [line 71](https://github.com/dotnet/samples/pull/1897/files#diff-a01210cbfe705b0745641383f463e05bR71). Makes sense to be consistent. Up to you.

snippets/core/system-text-json/csharp/Utf8ReaderFromFile.cs
```diff
@@ -10,7 +10,7 @@ public class Utf8ReaderFromFile
         private static readonly byte[] s_nameUtf8 = Encoding.UTF8.GetBytes("name");
         public static void Run()
         {
-            // Read as UTF-16 and transcode to UTF-8 to convert to a Span<byte>
+            // Read as UTF-16 and transcode to UTF-8 to convert to a ReadOnlySpan<byte>
```
ahsonkhan: Since we already mention it, maybe add the code snippet to strip the UTF-8 BOM.
```C#
public class Utf8ReaderFromFile
{
    private static readonly byte[] s_nameUtf8 = Encoding.UTF8.GetBytes("name");
    private static ReadOnlySpan<byte> Utf8Bom => new byte[] { 0xEF, 0xBB, 0xBF };

    public static void Run()
    {
        // Read as UTF-16 and transcode to UTF-8 to convert to a ReadOnlySpan<byte>
        //string fileName = @"C:\Users\ahkha\Desktop\jsonWithUtf8BOM.json";
        //string jsonString = File.ReadAllText(fileName);
        //ReadOnlySpan<byte> jsonReadOnlySpan = Encoding.UTF8.GetBytes(jsonString);

        // Or ReadAllBytes if the file encoding is UTF-8:
        string fileName = @"C:\Users\ahkha\Desktop\jsonWithUtf8BOM.json";
        ReadOnlySpan<byte> jsonReadOnlySpan = File.ReadAllBytes(fileName);

        // Read past the UTF-8 BOM bytes if one exists.
        if (jsonReadOnlySpan.StartsWith(Utf8Bom))
        {
            jsonReadOnlySpan = jsonReadOnlySpan.Slice(Utf8Bom.Length);
        }

        // Rest of the existing code
   }
}
```

PR #1900, opened 2020-01-09
Improve Process.Exited sample

MSDN-WhiteKnight:
## Summary

The current [Process.Exited](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.exited?view=netframework-4.8) example uses while loop to check when background thread sets the boolean field to wait for an event handler execution. This has a race condition, because the field is not marked volatile, and wastes CPU time. I suggested to make methods async and replace a loop with an await on TaskCompletionSource. Also it currently adds the amount of sleeps to calculate the elapsed time, but the simpler and more accurate way would be to simply subtract Process.StartTime from Process.ExitTime.

Replacement for https://github.com/dotnet/dotnet-api-docs/pull/3666 (that PR should be closed)

### DIFF
--- a/snippets/csharp/VS_Snippets_CLR_System/system.Diagnostics.Process.EnableExited/CS/processexitedevent.cs
+++ b/snippets/csharp/VS_Snippets_CLR_System/system.Diagnostics.Process.EnableExited/CS/processexitedevent.cs
@@ -1,19 +1,17 @@
 ﻿//<snippet1>
 using System;
 using System.Diagnostics;
-using System.Threading;
+using System.Threading.Tasks;
 
 class PrintProcessClass
 {
     private Process myProcess;
-    private int elapsedTime;
-    private bool eventHandled;
+    private TaskCompletionSource<bool> eventHandled;
 
     // Print a file with any known extension.
-    public void PrintDoc(string fileName)
+    public async Task PrintDoc(string fileName)
     {
-        elapsedTime = 0;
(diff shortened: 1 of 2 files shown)


NextTurn:
Why not `ManualResetEvent`?

MSDN-WhiteKnight:
> Why not `ManualResetEvent`?

Well, I was thinking about it. But reading the [remarks](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.exited?view=netframework-4.8#remarks) to this event:

>There are two ways of being notified when the associated process exits: synchronously and asynchronously. Synchronous notification means calling the WaitForExit method to block the current thread until the process exits. Asynchronous notification uses the Exited event, which allows the calling thread to continue execution in the meantime.

I realized that if I use 'ManualResetEvent.WaitOne' here, it will kind of defeat the purpose of the example, because it would be just reinventing the "synchronous" 'Process.WaitForExit', while the sample is supposed to show the different, "asynchronous" way. Then why not make it really asynchronous, in a modern way?

NextTurn:
> I realized that if I use `ManualResetEvent.WaitOne` here, it will kind of defeat the purpose of the example, because it would be just reinventing the "synchronous" `Process.WaitForExit`, while the sample is supposed to show the different, "asynchronous" way. Then why not make it really asynchronous, in a modern way?

Sounds great. It's a pity that VB can't have async `Main`.

PR #1945, opened 2020-01-31
Remove redundant null checks.

alimon808:
* The String.IsNullOrWhiteSpace method already checks for nulls.

### DIFF
--- a/snippets/csharp/tutorials/inheritance/base-and-derived.cs
+++ b/snippets/csharp/tutorials/inheritance/base-and-derived.cs
@@ -11,16 +11,12 @@ public abstract class Publication
 
    public Publication(string title, string publisher, PublicationType type)
    {
-      if (publisher == null)
-         throw new ArgumentNullException("The publisher cannot be null.");
-      else if (String.IsNullOrWhiteSpace(publisher))
-         throw new ArgumentException("The publisher cannot consist only of white space.");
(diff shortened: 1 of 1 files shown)


Youssef1313:
I think it's more suitable to throw ArgumentNullException if it's null, and ArgumentException if it's empty.

BillWagner:
Thanks for sparking this discussion @alimon808 and @Youssef1313 

I'm leaning toward accepting this PR, because for most uses combining the null check with an empty (and possibly white space) check is the best solution. 

It's also true that for wide use frameworks, separating those checks provides benefits for developers using those APIs.

I'll open this for discussion now.

alimon808:
To me it boils down to the intent of the constructor in this example and asking the following question: Is there a need to throw an ArgumentNull exception such that the consumer can handle it differently from an ArgumentException.  In the context of the [Inheritance in C#](https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/inheritance) documentation where the snippet is referenced, I didn't see such a need.  I don't know if this snippet is referenced in other documentations which could impact whether to merge these changes or not.  Can snippets from sample codes for one documentation be referenced in another?

BillWagner:
Thanks for the discussion everyone.

The consensus (both online and in other discussions) is to merge this.

I've approved the changes, and I'll :shipit: now.

PR #1949, opened 2020-01-31
Use string interpolation in snippet7

geekcyclist:
Snippet 7 reads:
`//Console.WriteLine("tra" + manyPhrases);`

In accordance with snippet6 it should read to be consistent:
`//Console.WriteLine($"tra{manyPhrases}");`

Unless part of the point of snippet7 is to also highlight performance differences between `System.String.Concat` (IL generated when using '+') and `System.String.Format` (IL generated when using string interpolation), in which case this pull request should be rejected.

## Summary

Describe your changes here.


BillWagner:
Thanks for creating this @geekcyclist 

I looked at where this sample is included, and it is intended to show the difference between string interpolation, concatenation, and using the StringBuilder type.

I'll close this.

PR #1960, opened 2020-02-04
use string.Create to programmatically construct content for a new string

Maoni0:
## Summary

This provides a sample using string.Create to programmatically construct content for a new string for https://github.com/dotnet/docs/pull/16908

### DIFF
--- a/snippets/csharp/how-to/strings/ModifyStrings.cs
+++ b/snippets/csharp/how-to/strings/ModifyStrings.cs
@@ -128,26 +128,23 @@ namespace HowToStrings
             // </Snippet6>
         }
 
-        private static void UnsafeSample()
+        private static void UsingStringCreate()
         {
             // <Snippet7>
-            unsafe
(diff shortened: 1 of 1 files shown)


snippets/csharp/how-to/strings/ModifyStrings.cs
```diff
@@ -128,26 +128,21 @@ private static void ReplaceCharArray()
...
-                Console.WriteLine(helloOne);
-                Console.WriteLine(helloTwo);
-            }
+	            strContent[0] = '0';
+	            strContent[1] = '1';
+	            for (int i = 0; i < charArray.Length; i++)
+	            {
+		            strContent[i + 2] = charArray[i];
+	            }
+            });
```
BillWagner: This is a great sample.

Can you add the following so readers can see the output:

```csharp
Console.WriteLine(result);
// output:
01abcd
```

PR #1988, opened 2020-02-10
Fix age calculation in deconstructing tuples sample

mj1856:
The age calculation in this sample was incorrect, and would fail for Feb 29th birth dates unless the current year is a leap year.

Though age calc is not the primary purpose of the example, it's still good to show it correctly to avoid copy/paste bugs.

The method used is the widely accepted approach from https://stackoverflow.com/questions/9/how-do-i-calculate-someones-age-in-c/1404#1404


tarekgh:
Thanks @mj1856 for fixing it.

PR #1989, opened 2020-02-10
Fix date calculation bug in MonthCalendarSelection example

mj1856:
The date manipulation being performed in these samples was incorrect, and would throw `ArgumentOutOfRangeException` in cases such as when the date was the first day of any month (because `today.Day - 1` would be `0`, which is invalid for the `DateTime` constructor).

The change corrects this by using `AddMonths` and `AddDays` methods.

Aside - it's unclear to me why `SelectionStart` and `SelectionEnd` need to be reduced by one day.  The docs don't spell that out, but the original samples did this and I needed to keep an `AddDays(-1)` to retain the same effect on the control.  Perhaps something here is 0-based instead of 1-based?


tarekgh:
Thanks @mj1856 for fixing it.

PR #2019, opened 2020-02-24
New code sample for dotnet/docs issue 12736

mssteele:
Part of a change to address issue:
https://github.com/dotnet/docs/issues/12736

My other change in the dotnet/docs repo contains the
corresponding documentation changes.
See PR: https://github.com/dotnet/docs/pull/17197

### DIFF
--- a/snippets/csharp/language-reference/keywords/is/is-var-pattern8.cs
+++ b/snippets/csharp/language-reference/keywords/is/is-var-pattern8.cs
@@ -14,23 +14,23 @@ class Program
                                     && factors.Contains(1)
                                     && factors.Contains(n));
 
-        foreach(int prime in primes)
+        foreach (int prime in primes)
         {
-            Console.WriteLine("Found prime: " + prime);
+            Console.WriteLine($"Found prime: {prime}");
         }
     }
 
     static IEnumerable<int> Factor(int number) 
     {
         int max = (int)Math.Sqrt(number);
-        for(int i = 1; i <= max; i++) 
+        for (int i = 1; i <= max; i++) 
         {
-            if(number % i == 0)
+            if (number % i == 0)
             {
                 yield return i;
-                if(i != number/i) 
+                if (i != number / i) 
                 {
-                    yield return number/i;
+                    yield return number / i;
                 }
             }
         }
--- a/snippets/csharp/language-reference/keywords/is/is-var-pattern8.cs
+++ b/snippets/csharp/language-reference/keywords/is/is-var-pattern8.cs
@@ -1,49 +1,42 @@
 ﻿// <Snippet8>
 using System;
+using System.Collections.Generic;
+using System.Linq;
 
 class Program
 {
     static void Main()
-   {
-      object[] items = { new Book("The Tempest"), new Person("John") };
-      foreach (var item in items) {
-        if (item is var obj)
-          Console.WriteLine($"Type: {obj.GetType().Name}, Value: {obj}"); 
-      }
-   }
-}
-
-class Book
-{
-    public Book(string title) 
     {
-       Title = title;    
+        int[] testSet = { 100271, 234335, 342439, 999683 };
+
+        var primes = testSet.Where(n => Factor(n).ToList() is var factors
+                                    && factors.Count == 2
+                                    && factors.Contains(1)
+                                    && factors.Contains(n));
+
+        foreach(int prime in primes)
+        {
+            Console.WriteLine("Found prime: " + prime);
+        }
     }
 
-    public string Title { get; set; }
-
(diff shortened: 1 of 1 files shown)


snippets/csharp/language-reference/keywords/is/is-var-pattern8.cs
```diff
@@ -1,49 +1,42 @@
...
     {
-       Title = title;    
-    }
+        int[] testSet = { 100271, 234335, 342439, 999683 };
 
-    public string Title { get; set; }
+        var primes = testSet.Where(n => Factor(n).ToList() is var factors
+                                    && factors.Count == 2
+                                    && factors.Contains(1)
+                                    && factors.Contains(n));
```
pkulikov: The last two checks are not necessary: `1` and `n` are always factors of `n`. So, it looks those two checks are only for the sake of the example of the need of the `factors` variable?

snippets/csharp/language-reference/keywords/is/is-var-pattern8.cs
```diff
@@ -1,49 +1,42 @@
...
 
-    public override string ToString()
-    {
-       return Title;
+        foreach(int prime in primes)
+        {
+            Console.WriteLine("Found prime: " + prime);
```
pkulikov:
```suggestion
            Console.WriteLine($"Found prime: {prime}");
```

snippets/csharp/language-reference/keywords/is/is-var-pattern8.cs
```diff
@@ -1,49 +1,42 @@
...
-   { get; set; }
-
-   public override string ToString()
-   {
-      return Name;
-   }
+    static IEnumerable<int> Factor(int number) 
+    {
+        int max = (int)Math.Sqrt(number);
+        for(int i = 1; i <= max; i++)
```
pkulikov:
```suggestion
        for (int i = 1; i <= max; i++) 
```

snippets/csharp/language-reference/keywords/is/is-var-pattern8.cs
```diff
@@ -1,49 +1,42 @@
...
-   public override string ToString()
-   {
-      return Name;
-   }
+    static IEnumerable<int> Factor(int number) 
+    {
+        int max = (int)Math.Sqrt(number);
+        for(int i = 1; i <= max; i++) 
+        {
+            if(number % i == 0)
```
pkulikov:
```suggestion
            if (number % i == 0)
```

snippets/csharp/language-reference/keywords/is/is-var-pattern8.cs
```diff
@@ -1,49 +1,42 @@
...
+            {
+                yield return i;
+                if(i != number/i)
```
pkulikov:
```suggestion
                if (i != number / i) 
```

snippets/csharp/language-reference/keywords/is/is-var-pattern8.cs
```diff
@@ -1,49 +1,42 @@
...
-    public string Title { get; set; }
+        var primes = testSet.Where(n => Factor(n).ToList() is var factors
+                                    && factors.Count == 2
+                                    && factors.Contains(1)
+                                    && factors.Contains(n));
 
-    public override string ToString()
-    {
-       return Title;
+        foreach(int prime in primes)
```
pkulikov:
```suggestion
        foreach (int prime in primes)
```

snippets/csharp/language-reference/keywords/is/is-var-pattern8.cs
```diff
@@ -1,49 +1,42 @@
...
+    {
+        int max = (int)Math.Sqrt(number);
+        for(int i = 1; i <= max; i++) 
+        {
+            if(number % i == 0)
+            {
+                yield return i;
+                if(i != number/i) 
+                {
+                    yield return number/i;
```
pkulikov:
```suggestion
                    yield return number / i;
```

snippets/csharp/language-reference/keywords/is/is-var-pattern8.cs
mssteele: Good point! But as you say removing them means we don't need the `factors` variable which is the point of the example. 
I could change the example to do something else with the factors, say find numbers with between 3 and 5 factors, if you think this example is a bit misleading?

BillWagner: I agree that the extra checks are unnecessary, but they do clarify the code.

I'm ready to approve this, pending the discussion on dotnet/docs#17197

PR #2028, opened 2020-02-26
Add SECURITY.md

terrajobst:
Add security notice

@BillWagner @mairaw can we fastrack this merge? @blowdart wants to have this file in all the repos that comprise the .NET Core product. I've already added it everywhere but due to branch policies, this one requires manual merge. Thanks!


terrajobst:
Thanks!

PR #2053, opened 2020-03-10
Add sample: Inspecting assembly contents using MetadataLoadContext

MSDN-WhiteKnight:
Contributes to <https://github.com/dotnet/docs/issues/17360>, https://github.com/dotnet/docs/issues/2729

Related docs PR: https://github.com/dotnet/docs/pull/17427


BillWagner:
I removed the team from the review list and assigned @gewarren because the related article PR is in her area.

pkulikov:
@BillWagner @Thraka should the new samples be added to the **dotnet/docs** repo as per dotnet/docs#17297?

MSDN-WhiteKnight:
> should the new samples be added to the **dotnet/docs** repo as per [dotnet/docs#17297](https://github.com/dotnet/docs/issues/17297)?

I think this one should stay there:

>The dotnet/samples repository will be the home of larger scenario-based samples that may be referenced from tutorials or other long-form explorations. 

Its not that large, but it is a scenario-based sample (a complete program for "How to" article). It also would be good to be downloadable via [sample browser](https://docs.microsoft.com/en-us/samples/browse/).

pkulikov:
> I think this one should stay there:

OK, but then, I'm afraid, you wouldn't be able to include the code snippet into the article like
`[!code-csharp[](~/samples/core/assembly/MetadataLoadContext/Program.cs)]` because the dotnet/docs repo is not aware of the dotnet/samples repo anymore. Somehow, the dotnet/docs PR should indicate that the snippet is to be taken from the dotnet/samples.

This is my current understanding of how code snippets are handled in the docs. @Thraka and @BillWagner know better.

MSDN-WhiteKnight:
I missed the meaning of the step "4. Remove the samples repo dependency". If that means cross-repo inclusion won't be possible, my PRs won't work as is. 

@BillWagner @gewarren I wanted the sample to be both included in the article and available in the Sample browser, but this seems not possible with planned samples changes (unless duplicating). How should i proceed here?

Thraka:
Duplicate.

You can tailor the snippet version to the context of the article, while the samples browser is complete and made for downloading.

As an example, the version for your article shouldn't just include the entire _program.cs_ file, it's a lot of code just plastered on the screen. At that point, the article and the sample readme are just about the same thing. It would be better for the article to break down the sample and explain the various pieces, each being a snippet hosted in the docs repo. At the end you would link to the sample.

core/assembly/MetadataLoadContext/Program.cs
```diff
@@ -0,0 +1,121 @@
...
+            {
+                Console.WriteLine("Usage: dotnet MetadataLoadContextSample.dll <assembly path>");
+                return 0;
+            }
+
+            string inputFile = args[0];
+
+            try
+            {
+                //get the array of runtime assemblies
```
gewarren: This is my first time reviewing a sample so perhaps @BillWagner or @Thraka can correct me here, but I think we need to capitalize comments and put a period on the end.

```suggestion
                // Get the array of runtime assemblies.
```

core/assembly/MetadataLoadContext/Program.cs
```diff
@@ -0,0 +1,121 @@
...
+                var mlc = new MetadataLoadContext(resolver);
+
+                using (mlc)
+                {
+                    //load assembly into MetadataLoadContext
+                    Assembly assembly = mlc.LoadFromAssemblyPath(inputFile);
+                    AssemblyName name = assembly.GetName();
+
+                    //print assembly attribute information
+                    Console.WriteLine(name.Name + " has following attributes: ");
```
gewarren: Use an interpolated string here?

core/assembly/MetadataLoadContext/Program.cs
```diff
@@ -0,0 +1,121 @@
...
+                        {
+                            //we are missing the required dependency assembly
+                            Console.WriteLine("Error: " + ex.Message);
+                        }
+                    }
+
+                    Console.WriteLine();
+
+                    //print assembly type information
+                    Console.WriteLine(name.Name + " contains following types: ");
```
gewarren: Interpolated string?

core/assembly/MetadataLoadContext/Program.cs
```diff
@@ -0,0 +1,121 @@
...
+                            {
+                                Console.Write("class ");
+                            }
+                            else if (t.IsValueType)
+                            {
+                                if (String.Equals(baseType?.FullName, "System.Enum", StringComparison.InvariantCulture))
+                                {
+                                    Console.Write("enum ");
+                                }
+                                else
```
gewarren: Is struct really the only other option? Wouldn't it better to explicitly match it for future proofing?

core/assembly/MetadataLoadContext/Program.cs
```diff
@@ -0,0 +1,121 @@
...
+                    foreach (CustomAttributeData attr in assembly.GetCustomAttributesData())
+                    {
+                        try
+                        {
+                            Console.WriteLine(attr.AttributeType);
+                        }
+                        catch (FileNotFoundException ex)
+                        {
+                            //we are missing the required dependency assembly
+                            Console.WriteLine("Error: " + ex.Message);
```
gewarren:
```suggestion
                            Console.WriteLine("Error getting attribute type: " + ex.Message);
```

core/assembly/MetadataLoadContext/readme.md
```diff
@@ -0,0 +1,22 @@
+---
+languages:
+- csharp
+products:
+- dotnet-core
+page_type: sample
+name: "Inspect assembly contents using MetadataLoadContext"
+urlFragment: "inspect-assembly-contents-using-metadataloadcontext"
+description: "A sample console application that shows how to use MetadataLoadContext to load an assembly for inspection purposes"
```
gewarren:
```suggestion
description: "A sample console application that shows how to use MetadataLoadContext to load an assembly for inspection purposes."
```

core/assembly/MetadataLoadContext/readme.md
```diff
@@ -0,0 +1,22 @@
...
+---
+# Inspect assembly contents using MetadataLoadContext
+
+This sample is a .NET Core console application that loads the specified assembly and prints its custom attributes and defined types. It demonstrates how to use the [System.Reflection.MetadataLoadContext](https://www.nuget.org/packages/System.Reflection.MetadataLoadContext) API to load an assembly for inspection purposes.
```
gewarren:
```suggestion
This sample is a .NET Core console application that loads a specified assembly and prints its custom attributes and defined types. It demonstrates how to use the [System.Reflection.MetadataLoadContext](https://www.nuget.org/packages/System.Reflection.MetadataLoadContext) API to load an assembly for inspection purposes.
```

core/assembly/MetadataLoadContext/readme.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Build and run the sample
+
+After downloading the source, open the console in the sample directory and use following commands:
```
gewarren:
```suggestion
After downloading the source, open a command prompt in the sample directory and enter the following commands:
```

core/assembly/MetadataLoadContext/Program.cs
MSDN-WhiteKnight: General [comment conventions](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions#commenting-conventions) recommend the same, so i'll apply this.

core/assembly/MetadataLoadContext/Program.cs
MSDN-WhiteKnight: It is the only other option currently (technically built-in value types like `int` aren't structures by ECMA spec, but they are still declared as structs in corelib metadata). But i'll add a check that the struct is inherited from ValueType directly to future-proof against possible other special value types.

gewarren:
@MSDN-WhiteKnight Thanks so much for adding this sample.

PR #2793, opened 2020-03-16
Add stack sampling profiler sample

davmason:
## Add stack sampling profiler sample

Moves https://github.com/dotnet/runtime/pull/32988 to this repo. The only changed files are adding readmes in the appropriate folders.


core/profiling/stacksampling/src/sampler.cpp
```diff
@@ -0,0 +1,380 @@
...
+
+    pProfInfo->InitializeCurrentThread();
+
+    while (true)
+    {
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
+
+        s_waitEvent.Wait();
+
+        if (!parent->IsRuntimeExecutingManagedCode())
```
noahfalk: I'd recommend adding some comments around this to clarify:
1. This isn't required for correctness, it was convenient for demonstration purposes
2. The implementation of this call isn't a precise way to determine if managed code is running because the runtime can always invoke pre-compiled code before jitting. However for a sample it was an easy to implement heuristic.

davmason: Thanks!

PR #2885, opened 2020-05-12
Publish for sample copy (prior to delete)

Thraka:


BillWagner:
@Thraka I did the regular publish this morning, so I think this can be safely closed.

Thraka:
@BillWagner How do I know it worked? I'm also not sure if there is a problem with the sample or not. The link doesn't seem to work: https://docs.microsoft.com/en-us/samples/dotnet/samples/windowsforms-formatting-utility-cs

PR #3598, opened 2020-05-15
Add datetime.parse and string.format C# samples

gewarren:
These samples were accidentally deleted in #2064.


BillWagner:
I agree that we should restore these. But, should they now be in the dotnet-api-docs repo instead?

gewarren:
> I agree that we should restore these. But, should they now be in the dotnet-api-docs repo instead?

I'll do whatever is recommended by you and @thraka. These snippets already exist in the API docs repo, but in a slightly different format. For example, [this snippet in the API docs repo](https://github.com/dotnet/dotnet-api-docs/tree/master/samples/snippets/csharp/VS_Snippets_CLR_System/system.String.Format/cs/) is encompassed [in this file that's part of the sample project](https://github.com/dotnet/samples/blob/8f727f29eb1dcdbb3c79479f2fccbce30a7dea3f/csharp/api/string.format/q-a.cs). So in the API repo, they exist as standalone files without a project file. In this format that I need to resurrect somewhere, the snippet tags aren't included and various snippets are combined into a runnable program with command line UI.

gewarren:
@Thraka Do you think I should move this into the API repo instead, even though there will be some duplication in the code?

Thraka:
This looks complete and contained, it's designed as a sample, not as a snippet. I think we should restore it and convert it to a true sample. I also noticed there is a `datetime.parse` sample in the same parent folder.

gewarren:
@BillWagner or @Thraka Does this look good? Can you approve if so?

PR #3703, opened 2020-07-06
Add CI validation

adegeo:
I just copied the workflow and dependencies from the docs repo.


adegeo:
Yeah. Looks like GitHub doesn't include those by default. Whoops

PR #3711, opened 2020-07-11
Re-generate dependabot.yml

Youssef1313:
Re-generated by https://github.com/BillWagner/dependadotnet/tree/version-2-syntax

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,6 +2,11 @@
 version: 2
 updates:
 - package-ecosystem: nuget
+  directory: "/azure/sdk-identity-resources-storage" # AzureIdentityStorageExample.csproj
+  schedule:
+    interval: daily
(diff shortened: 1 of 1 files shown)


BillWagner:
Waiting on this PR. We have a question to he dependabot team regarding configuration and navigating child folders. Pending that outcome, the configuration may get much simpler.

IEvangelist:
Here is the ask: https://github.com/dependabot/dependabot-core/issues/2302

IEvangelist:
> Waiting on this PR. We have a question to he dependabot team regarding configuration and navigating child folders. Pending that outcome, the configuration may get much simpler.

I heard back [here](https://github.com/dependabot/dependabot-core/issues/2302#issuecomment-658050362), and have confirmed that the configuration is not that simple - but they're looking into ways to make that happen. The next several months are dedicated to integrating with GitHub.

PR #3742, opened 2020-07-18
Make method name PascalCase

Youssef1313:


adegeo:
We don't need to change private members. I would probably just leave this alone since the author probably did it intentionally?

Youssef1313:
Wondering why this could be done intentionally?
A small side question: where this sample actually used?

adegeo:
"samples" here are supposed to be downloaded and run as examples, not referred to via the docs and "snippets" but to answer your question, it's a typical styling to not use PascalCase on private members.

Youssef1313:
I personally didn't see the camelCase for private methods styling before, but I'm okay too with closing if you prefer to keep it camelCase.

adegeo:
I'd prefer not to change anything that isn't critical 😄 If someone is redoing the whole sample and wants to use a different style for private, that's fine.

Thanks though 😄 😄

adegeo:
Originally the framework design guidelines don't cover private. Private is only to you, the programmer, and not to any publish consumers of your app/library, so there aren't any set guidelines.

PR #3758, opened 2020-07-28
add snippets 5000 config file

BillWagner:
Roslyn samples must be built using Visual Studio.


adegeo:
That shouldn't matter. The snippets file is only analyzed with a result returned from the locate projects. So when it goes to compile the project it 1st checks is there a snippets file in the same folder or not.

PR #3819, opened 2020-08-11
Added sync, serial-async, and parallel-async examples

IEvangelist:
## Summary

Added sync, serial-async, and parallel-async examples.

Related to:


async/async-and-await/cs/ParallelAsyncExample/MainWindow.xaml.cs
```diff
@@ -0,0 +1,95 @@
...
+using System.Net.Http;
+using System.Threading.Tasks;
+using System.Windows;
+using System.Windows.Threading;
+
+namespace ParallelAsyncExample
+{
+    public partial class MainWindow : Window
+    {
+        readonly HttpClient _client = new HttpClient { MaxResponseContentBufferSize = 1_000_000 };
```
Youssef1313, line 14:
```suggestion
        private readonly HttpClient _client = new HttpClient { MaxResponseContentBufferSize = 1_000_000 };
```

async/async-and-await/cs/ParallelAsyncExample/MainWindow.xaml.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        readonly IEnumerable<string> _urlList = new string[]
```
Youssef1313, line 16:
```suggestion
        private readonly IEnumerable<string> _urlList = new string[]
```

async/async-and-await/cs/ParallelAsyncExample/MainWindow.xaml.cs
```diff
@@ -0,0 +1,95 @@
...
+            "https://docs.microsoft.com/surface",
+            "https://docs.microsoft.com/xamarin",
+            "https://docs.microsoft.com/azure/devops",
+            "https://docs.microsoft.com/system-center",
+            "https://docs.microsoft.com/graph",
+            "https://docs.microsoft.com/education",
+            "https://docs.microsoft.com/gaming",
+        };
+
+        void OnStartButtonClick(object sender, RoutedEventArgs e)
```
Youssef1313, line 39: Access modifiers are missing to all methods

async/async-and-await/cs/ParallelAsyncExample/MainWindow.xaml.cs
```diff
@@ -0,0 +1,95 @@
...
+            {
+                _resultsTextBox.Text += $"\nControl returned to {nameof(OnStartButtonClick)}.";
+                _startButton.IsEnabled = true;
+            });
+        }
+
+        async Task SumPageSizesAsync()
+        {
+            var stopwatch = new Stopwatch();
+            stopwatch.Start();
```
Youssef1313, line 60: nit: You can use the Stopwatch.StartNew static method.

async/async-and-await/cs/ParallelAsyncExample/MainWindow.xaml.cs
```diff
@@ -0,0 +1,95 @@
...
+
+            return byteArray.Length;
+        }
+
+        Task DisplayResultsAsync(string url, byte[] content) =>
+            Dispatcher.BeginInvoke(() =>
+                _resultsTextBox.Text += $"{url,-60} {content.Length,10:#,#}\n")
+                      .Task;
+
+        protected override void OnClosed(EventArgs e) => _client?.Dispose();
```
Youssef1313, line 93: It doesn't seem possible that `_client` can be null. Is the check for null just defensive?

async/async-and-await/cs/readme.md
```diff
@@ -5,42 +5,35 @@ products:
 - dotnet-core
```
Youssef1313, line 5: This sample will appear under the ".NET Core" node in the samples browser. Currently, there is no option in the samples browser for .NET 5.0 and later, so nothing to do in the PR. Can you submit an internal request for that?

Here is an image of the current available options under the .NET node.

image

async/async-and-await/cs/ParallelAsyncExample/MainWindow.xaml.cs
IEvangelist: They're not missing, they were intentionally omitted. I prefer to omit default access modifiers as I believe it makes it easier to read, and they're redundant. But, I added them.

PR #3823, opened 2020-08-12
Fix sample build for framework project

BillWagner:
This should fix the build errors in #3813


BillWagner:
Hey @adegeo 

This looks like a bug in the powershell script. Or in how we're calling MS Build for a framework solution.

PR #3865, opened 2020-08-25
Added GrapeCity FlexGrid Showcase sample

gc-deniszmitrichenko:
## Summary

Added GrapeCity FlexGrid Showcase sample.
@OliaG told us that we can submit a PR with our sample here.


gc-deniszmitrichenko:
Thank you for review, @adegeo!
I'll make changes and submit another PR.

PR #3868, opened 2020-08-26
Added GrapeCity FlexGrid Showcase sample

gc-deniszmitrichenko:
## Summary

Added GrapeCity FlexGrid Showcase sample


windowsforms/FlexGridShowcaseDemo/cs/README.md
```diff
@@ -0,0 +1,49 @@
+﻿---
```
adegeo, line 1: This file has a UTF-16 (or something) code at the start (invisible here) I think this needs to be converted to UTF-8. You can see it if you use [edit file](https://github.com/gc-deniszmitrichenko/samples/edit/master/windowsforms/FlexGridShowcaseDemo/cs/README.md?pr=%2Fdotnet%2Fsamples%2Fpull%2F3868).

PR #3906, opened 2020-09-15
Snippets config: Remove CPP triggers; Update to .NET 5 RC 1

adegeo:


BillWagner (approved):
This LGTM. Also same question on moving directly to RC1.

PR #3909, opened 2020-09-18
Update HostingWithHostFxr to 3.1

vitek-karas:
Also small changes to the readme

### DIFF
--- a/core/hosting/HostWithHostFxr/readme.md
+++ b/core/hosting/HostWithHostFxr/readme.md
@@ -1,6 +1,6 @@
 # .NET Core Hosting Sample
 
-This project demonstrates a way for a native process to host .NET Core using the `nethost` and `hostfxr` libraries. Documentation on the `nethost` and `hostfxr` APIs can be found [here](https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/native-hosting.md).
+This project demonstrates a way for a native process to host .NET Core using the `nethost` and `hostfxr` libraries. Documentation on the `nethost` and `hostfxr` APIs can be found [here](https://github.com/dotnet/runtime/blob/master/docs/design/features/native-hosting.md).
 
 ## Key Features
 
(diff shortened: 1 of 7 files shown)


vitek-karas:
The failure occurs because this sample requires native compilers to be available (C++ compiler) on the machine and for the build to run from Visual Studio Developer environment (so that the compilers are accessible).

I guess this won't work in the CI build system - is there a way to tell the CI to not build a specific project?

adegeo:
Add a file named snippets.5000.json as peer to the solution file you want to compile with visual studio and put this inside of it:

```json
{
    "host": "visualstudio"
}

```

This forces the build system to run the vs dev prompt and msbuild instead of dotnet.

elinor-fung (commented):
The other complication for the build is that, since we use the bitness of `dotnet` to find the corresponding `nethost` lib, we need the bitness of the native compiler to match that of `dotnet` - and the normal VS developer command prompt has the 32-bit native compiler.

We could also try switching it to find the compiler automatically on Windows - similar to what the `IDynamicCastableInterface` (https://github.com/dotnet/samples/pull/3744/commits/82748ddeb041b936c45b94eb620a8548feabcf70) and p/invoke marshalling (https://github.com/dotnet/samples/commit/3bbeef4a3ed0ee95cd0da3465cb3d91a393731cb) samples do.

adegeo:
How much config changes do you expect users to do when they download the sample?

vitek-karas:
Ideally none - it should just work.
I'm following @elinor-fung advice and porting over the native build from the marshalling sample. That should be buildable without issues and without environment (basically the same as marshalling sample)

vitek-karas:
Finally reworked it so that it should not need any special treatment. Works both from command line and VS, also works on Linux.
Ready for review.

BillWagner (approved):
This LGTM @vitek-karas 

I'll :shipit: now.

PR #3917, opened 2020-10-01
Move "dotnet restore" note

samdanae:
## Summary

There is no need for this sentence any more since the previous code block does not mention `dotnet restore`. Moved the "see note" link to the next sentence.


BillWagner (approved):
I like this wording @samdanae 

I've approved this PR, and I'll :shipit: now.

You should see the changes on the live site in the next day or so, on our regular publishing cycle.

Thanks for making your first contribution to docs! 🎆

PR #3921, opened 2020-10-05
Update pattern matching tutorial sample for C# 9

BillWagner:
contributes to dotnet/docs#20594

Add one example for or patterns and relational patterns. Also, show a pre-pattern algorithm for comparison.

Tagging @adegeo for review since he recently updated this as well.


IEvangelist (approved):
Looks good to me, :shipit: when you're ready.

PR #4006, opened 2020-11-21
Update Benchmark to reflect optimisations

LokiMidgard:
## Summary

The [docs](https://docs.microsoft.com/en-us/dotnet/csharp/write-safe-efficient-code#avoid-mutable-structs-as-an-in-argument) say that using `in` modifier with non `readonly struct`s can have performance issues. It references this file that should demonstrate this. However when I tested it on *.Net 5* I didn't see any difference.
I also got the warning that some calls could not be distinguished from empty methods. 

This I made following changes:
* _add a return type to benchmark methods_  
  According to the [documentation of Benchmark .Net](https://benchmarkdotnet.org/articles/guides/good-practices.html#avoid-dead-code-elimination) this prevents dead
  code elimination.
* _refactored auto properties to full properties_  
  It seems since the benchmark was written some optimization was done to auto properties, so they are marked as readonly. This no longer demonstrates the effect desired
* _Add a third `MutableReadonly` struct was added_  
  To show that Mutable `structs` do not automatically decrease performance. Which is also important since having a big `readonly struct` where one field needs often to be
  changed, would result in a decrease in performance.

## Resulting Benchmark 

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19041.630 (2004/?/20H1)
Intel Core i7-8650U CPU 1.90GHz (Kaby Lake R), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=5.0.100
  [Host]     : .NET Core 5.0.0 (CoreCLR 5.0.20.51904, CoreFX 5.0.20.51904), X64 RyuJIT
  DefaultJob : .NET Core 5.0.0 (CoreCLR 5.0.20.51904, CoreFX 5.0.20.51904), X64 RyuJIT

|                      Method |      Mean |     Error |    StdDev |
|---------------------------- |----------:|----------:|----------:|
|          ImmutableAddByType | 3.1876 ns | 0.0769 ns | 0.0681 ns |
|       ImmutableAddByRefType | 0.2507 ns | 0.0229 ns | 0.0191 ns |
|            MutableAddByType | 3.1468 ns | 0.0454 ns | 0.0402 ns |
|         MutableAddByRefType | 7.7434 ns | 0.1459 ns | 0.1293 ns |
|    MutableReadOnlyAddByType | 3.5987 ns | 0.0570 ns | 0.0533 ns |
| MutableReadonlyAddByRefType | 0.2628 ns | 0.0317 ns | 0.0281 ns |

### DIFF
--- a/csharp/safe-efficient-code/benchmark/Program.cs
+++ b/csharp/safe-efficient-code/benchmark/Program.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using BenchmarkDotNet.Attributes;
 using BenchmarkDotNet.Running;
 
@@ -36,9 +36,9 @@ namespace benchmark
 
     public struct MyMutableStruct
     {
-        public double X { get; }
-        public double Y { get; }
-        public double Z { get; }
+        public double X { get => x; set => x = value; }
+        public double Y { get => y; set => y = value; }
+        public double Z { get => z; set => z = value; }
 
         private double a;
         private double b;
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
Thanks for making these changes @lokiMidgard We appreciate it.

I've reviewed the code, and pulled this PR and ran the sample. I'll :shipit: now.

PR #4042, opened 2020-12-10
Update TaskDialog sample README.md

RussKie:
Provide a correct link to the .NET docs instead of Win32 docs.


merriemcgaw (approved):
Works for me

PR #4044, opened 2020-12-11
Add support for custom hosts

adegeo:
## Summary

- Add logic to check for "custom" host in snippets.5000.json file.
- "dotnet" host will use the default value.
- Throw error if snippets.5000.json file present but nothing used from it.


BillWagner (approved):
This LGTM @adegeo 

You can :shipit: when ready

PR #4067, opened 2020-12-20
Fix TaskDialogDemo compilation

EatonZ:
1. Removes WindowsDesktop part of the SDK. [Reference](https://docs.microsoft.com/en-us/dotnet/core/compatibility/windows-forms/5.0/sdk-and-target-framework-change)
2. Fixes compilation error - net5.0-windows is the correct TargetFramework. Won't compile otherwise.

Might want to check the other samples too.😉


RussKie (approved):
Thank you

EatonZ:
Wow. Fastest approval/merge ever.😁

RussKie:
I happen to open the email :)

RussKie:
Do you think you could update the VB sample as well? https://github.com/dotnet/samples/blob/master/windowsforms/TaskDialogDemo/vb/TaskDialogDemo.vbproj

PR #4251, opened 2021-02-01
add solution files

BillWagner:
If a solution file changes, we should build some code.

Found while working with @scottaddie on installing this in the asp.net core docs repo.


IEvangelist (approved):
Looks good to me, :shipit: when you're ready. Same comment as with donet/docs, this was intentionally added to the *.gitignore* file.

PR #4328, opened 2021-02-15
Fix dll path in Unloading sample

jc1001:
## Summary

The changes made in https://github.com/dotnet/samples/pull/4256 broke the unloading sample.  It was still trying to load the dll from netcoreapp3.0, not netcoreapp3.1.


BillWagner (approved):
Thanks for making this change @jc1001 

I'll :shipit: now.

PR #4330, opened 2021-02-16
Finish getstarted tutorials

BillWagner:
contributes to dotnet/docs#19715

related to dotnet/docs#22820

Update our introductory sample code to C# 9.0

These samples are updated to use top-level statements. We want to introduce beginners to top-level statements early in their learning.


IEvangelist (approved):
Looks good to me, :shipit: when you're ready.

PR #4336, opened 2021-02-17
console-teleprompter: fix case sensitive bug

ccouzens:
The program has a file in the local directory called sampleQuotes.txt
(lowercase s at the beginning).

The program tried to read it as SampleQuotes.txt (uppercase s at the
beginning). On case sensitive filesystems such as on Linux, the file
couldn't be found.

I didn't rename the file, because it is referred to by
https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/console-teleprompter.

Instead, I fixed the file name that's requested in the program.


BillWagner (approved):
Thanks for making this change @ccouzens 

I've reviewed the changes and I'll :shipit: now.

Issue #4378, opened 2021-03-02
MakeConst sample needs to be updated


Youssef1313:
It should have changes parallel to https://github.com/dotnet/docs/pull/23066

PR #4379, opened 2021-03-02
Update roslyn-sdk analyzer sample

Youssef1313:
Contributes to #4378


csharp/roslyn-sdk/Tutorials/MakeConst/MakeConst.Test/Verifiers/CSharpAnalyzerVerifier`1+Test.cs
```diff
@@ -0,0 +1,26 @@
+﻿using Microsoft.CodeAnalysis.CSharp.Testing;
```
BillWagner, line 1: I'm curious.

This file, and several others, have strange name (the + and the ` character.)  Is that intentional? And why?

BillWagner (commented):
The code looks great @Youssef1313 

I had one comment on the filenames. Once we resolve that, this is great.

csharp/roslyn-sdk/Tutorials/MakeConst/MakeConst.Test/Verifiers/CSharpAnalyzerVerifier`1+Test.cs
Youssef1313: @BillWagner This is what the default template has. The naming for this is made the same as the symbol metadata name.

BillWagner (approved):
This LGTM.

Thanks for all this work @Youssef1313 It's really good work.

PR #4401, opened 2021-03-08
Delete MakeConst from dependabot config

Youssef1313:
Updating dependencies for this analyzer sample isn't so good as it new CodeAnalysis packages will require higher version of Visual Studio and/or SDK.


BillWagner (approved):
Thanks @Youssef1313 

I'll :shipit: now.

PR #4435, opened 2021-03-16
Create dotnet-code-metrics.yml

IEvangelist:
## Summary

Adding a code metrics workflow to test the sample .NET GitHub action.

Related to dotnet/docs#23117

### DIFF
--- a/.github/workflows/dotnet-code-metrics.yml
+++ b/.github/workflows/dotnet-code-metrics.yml
@@ -33,4 +33,9 @@ jobs:
         name: ${{ github.repository }}
         branch: ${{ github.ref }}
         dir: ${{ './github-actions/DotNet.GitHubAction' }}
-        
+      
+    - name: Create pull request
+      uses: peter-evans/create-pull-request@v3.4.1
+      with:
+        title: 'Update CODE_METRICS.md'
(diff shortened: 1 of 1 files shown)


.github/workflows/dotnet-code-metrics.yml
```diff
@@ -0,0 +1,36 @@
+name: '.NET code metrics'
+
+on:
+  push:
```
BillWagner, line 4: Do you want to add running this on PR open & commit? (Open question, I can argue both sides. I'd like to hear your thoughts.

BillWagner (approved):
This LGTM @IEvangelist 

I'm really curious to see what it says about our code. 

I had one question, mostly out of pure curiosity. You can :shipit: when ready.

.github/workflows/dotnet-code-metrics.yml
IEvangelist: The GitHub Action that it will be calling into, evaluates code metrics for a targeted `dir`. It will not do code metrics on all the samples, that would be a LOT. The action creates a PR to maintain a CODE_METRICS.md file. I like `on push` here, since it is final. I don't think we'd want a PR on ever rev of a commit.

PR #4431, opened 2021-03-12
.NET GitHub action

IEvangelist:
## Summary

A Docker containerized .NET app, that serves as a GitHub action. When ran will analyze C# and VB projects for code metrics, reporting:

- Cyclomatic complexity
- Maintainability index
- Class coupling
- Depth of inheritance
- Exact number of lines of code
- Approximated number of lines of executable code

It writes a *CODE_METRICS.md* file at the root directory given, and will enable GitHub action workflow composition to create PRs. In other words, consuming this GitHub action with the create-pull-request action in a .NET repo will provide automatic PRs to maintain the code metrics as a markdown file.

Ignore the `DotNet.CodeAnalysis` project in the review it is mostly borrowed source, instead focus on the `DotNet.GitHubAction` bits.

Contributes to dotnet/docs#23117

PR #4446, opened 2021-03-17
We need to persist the build-env layer from the SDK

IEvangelist:
## Summary

We need to persist the build-env layer from the SDK, so that MSBuild bits are discoverable. Instead of building the final image layer as the runtime, instead leave it as the SDK (which includes the runtime), but also has the msbuild bits...

Related to #4431 and #4435


BillWagner (approved):
This LGTM @IEvangelist 

Go ahead and :shipit:

PR #4447, opened 2021-03-17
Added workflow composition YML to SLN for easier edits, fix two build…

IEvangelist:
## Summary

- Added workflow composition YML to SLN for easier edits
- Fix two build warnings in container
- Fix path COPY cmds
- Add `paths` limit, to ensure the workflow only runs when needed

Related to #4446


BillWagner (approved):
This LGTM @IEvangelist 

You can :shipit: when ready.

PR #4448, opened 2021-03-17
Re-layer anew from the .NET SDK, to avoid COPY cmd issues

IEvangelist:
## Summary

Re-layer anew from the .NET SDK, to avoid COPY cmd issues

Related to #4447


BillWagner (approved):
This LGTM.

PR #4449, opened 2021-03-17
No longer need GH-token, but still want to show env var passing. Also…

IEvangelist:
## Summary

No longer need GH-token, but still want to show env var passing. Also, want to show conditional steps and output params.

Related to #4448


BillWagner (approved):
This LGTM. WE're getting there.

PR #4453, opened 2021-03-17
A few more updates to the tutorial bits and the code metrics MD file

IEvangelist:
## Summary

- Avoid triggering on self PRs
- Added a comment to the MD file to disable/capture, and then restore MD lint rules - for the file that is created.

Related to #4450


BillWagner (approved):
This LGTM @IEvangelist 

You can :shipit: when ready.

PR #4456, opened 2021-03-17
There was an issue with the generated line number

IEvangelist:
## Summary

There was an issue with the generated line number, this fixes that.

Related to #4451


BillWagner (approved):
This LGTM @IEvangelist 

You can :shipit: when ready.

PR #4492, opened 2021-04-01
Add status-checker GitHub Action

gewarren:


.github/actions/status-checker/.eslintignore
```diff
@@ -0,0 +1,4 @@
+dist/
```
gewarren, line 1: Not sure if I should check this in?

BillWagner (approved):
This is ready to deploy and start testing.

Let's :shipit: and see how it works.

PR #4497, opened 2021-04-02
Fix array length oops

gewarren:


BillWagner (approved):
Let's :shipit:

.github/actions/status-checker/src/status-checker.ts
```diff
@@ -12,13 +12,13 @@ export async function checkStatus(token: string) {
 
   if (prNumber) {
 
-    const { data: pullCommits } = await octokit.rest.pulls.listCommits({
+    const { data: pullCommits } = await octokit.pulls.listCommits({
```
gewarren, line 15: @BillWagner The second commit just removes "rest" here. That seems to be the normal way to call this API.

PR #4532, opened 2021-04-08
Fix if/else logic

gewarren:


.github/actions/status-checker/src/status-checker.ts
```diff
@@ -40,16 +41,18 @@ export async function checkStatus(token: string) {
         }
       }
 
-      if (buildStatus != null && buildStatus.state == 'pending') {
-        console.log("Found OPS status check but it's still pending.")
-        // Sleep for 10 seconds.
-        await wait(10000);
-        continue;
-      }
-      else {
```
gewarren, line 49: buildStatus could still be null here and we'd break, which is wrong.

adegeo (approved):

PR #4536, opened 2021-04-12
Added null-coalesce with the emoji, again can't assume  will succeed.…

IEvangelist:
- Added null-coalesce with the emoji, can't assume `.First()` will succeed Use `.FirstOrDefault()` instead.


BillWagner (approved):
This LGTM @IEvangelist 

You can :shipit:

PR #4543, opened 2021-04-21
Some cleanup

gewarren:
- allow pull_request and pull_request_target events
- add const for github.context.payload


.github/actions/status-checker/src/status-checker.ts
```diff
@@ -7,17 +7,18 @@ export async function checkStatus(token: string) {
...
-  if (github.context.eventName === 'pull_request_target' && github.context.payload?.action) {
-    const prNumber = github.context.payload.number;
+  if (['pull_request', 'pull_request_target'].includes(github.context.eventName) && payload?.action) {
+    const prNumber = payload.number;
     console.log({ prNumber });
 
     let sha;
-    if (github.context.payload.action === 'synchronize') {
-      sha = github.context.payload.after;
+    if (payload.action === 'synchronize') {
```
IEvangelist, line 17: It might be better to use a `switch` rather than `if/else if/else`, here is an example [TypeScript playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAggxsAlgewHZQLxQOQGcSpwAWATmogF4TZQA+OykqEAJjfdiRIxM2wNwBYAFChIUAAoBXADYyAShACOUiLmCYoAbxFQoRCAEMWALm269UXEUNn1JRKgDmFgL4j3wkWOgTDIGWRjTR1hPUMAM2AIEgB+O2AHZwAaC0MEFFQzeCQ0CzBZGQB9LhU1YHjJQsUy9TooKVQWCAjHVg8hL2E4NDqwf0DjMz8AoJYQtIy0MzwCYjJUSmpUsKgCuRLlVXUzUMt9I1NzVf3rWxxDACM4AEYAJgBmbAs9T1eOkREZCA0zhKSnPVGs1WnxOrgAO6IYDEKAACn6o2MADp0rlUABKY77OCGXDQWaEUjkKjYEwvSxnTSIwYsVFRGKdfZ6S5cQwAayZllx+IYTFYZIpPIJXB4fEFJ0pNmpAzGyPWxVK2wqyIMKLOXP2rKMnIpIMMsmA5Mleh6qFwyG+yMCTjh2AAqswAB6QBCsNay4JozLI7AYzWWbUcrmuIA):
```typescript
switch (payload.action) {
    case 'synchronize':
      sha = payload.after;
      break;
    case 'opened':
    case 'reopened':
      sha = payload.pull_request?.head.sha;
      break;
    default:
      console.log('Unexpected payload action.');
      break;
  }
```

IEvangelist (approved):
I have a suggestion, but this will still work.

.github/actions/status-checker/src/status-checker.ts
gewarren: I'm actually going to try to remove it altogether to simplify. I think I can still get the commit SHA from pull_request.head.sha, even if it's a synchronize action.

PR #4583, opened 2021-05-05
See if we can get download zip added to matching game samples

adegeo:
Added front matter to see if it enables the **Download ZIP** button.

dotnet/docs#23380

### DIFF
--- a/windowsforms/matching-game/net5-windows/cs/readme.md
+++ b/windowsforms/matching-game/net5-windows/cs/readme.md
@@ -8,7 +8,10 @@ products:
 page_type: sample
 name: "Matching Game Sample (.NET 5 C#)"
 urlFragment: "matching-game-net-csharp"
-description: "A simple matching game "
+description: "A simple matching game written for Windows Forms"
+extendedZipContent:
+- path: null
+- target: null
 ---
 # Matching game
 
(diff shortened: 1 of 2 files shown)


tdykstra (approved):
Let me know if this works, I have a sample that is missing zip download too.

Youssef1313:
@adegeo This doesn't seem to work.

This is a sample that has the button:

https://docs.microsoft.com/en-us/samples/azure/azure-sdk-for-net/share-link/

It adds real files to `extendedZipContent`

https://raw.githubusercontent.com/Azure/azure-sdk-for-net/main/sdk/keyvault/samples/sharelink/README.md

We may want to try that by including [LICENSE](https://github.com/dotnet/samples/blob/main/LICENSE) for example.

gewarren:
They don't generate ZIP downloads by default, but last I heard, even though the [contributor guide](https://review.docs.microsoft.com/en-us/help/contribute/samples/process/onboarding?branch=master#supported-metadata-fields-for-readmemd) says that `extendedZipContent` is "deprecated", is that if you specify a non-null value for it, a ZIP file will be generated. See https://github.com/dotnet/samples/issues/4723 for more discussion.

PR #4585, opened 2021-05-11
Use state of current status check

gewarren:
Set state of current status check instead of inserting an additional status check, which hangs around on `reopened` and shows the old status.


BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when ready.

gewarren:
Thank you for reviewing, Bill!

PR #4593, opened 2021-05-19
use nuget packages in samples instead of expecting visual studio to be installed

jmarolf:
## Summary

Describe your changes here.

### DIFF
--- a/csharp/roslyn-sdk/Tutorials/MakeConst/MakeConst.Vsix/MakeConst.Vsix.csproj
+++ b/csharp/roslyn-sdk/Tutorials/MakeConst/MakeConst.Vsix/MakeConst.Vsix.csproj
@@ -19,7 +19,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.VSSDK.BuildTools" Version="15.1.192" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.VisualStudio.SDK" Version="16.0.206" ExcludeAssets="runtime" />
+    <PackageReference Include="Microsoft.VSSDK.BuildTools" Version="17.0.1597" />
   </ItemGroup>
 
   <PropertyGroup>
@@ -35,8 +36,6 @@
 
 
   <Import Sdk="Microsoft.NET.Sdk" Project="Sdk.targets" />
-
-  <Import Project="$(VSToolsPath)\VSSDK\Microsoft.VsSDK.targets" Condition="Exists('$(VSToolsPath)\VSSDK\Microsoft.VsSDK.targets')" />
 
   <ItemGroup>
     <!-- https://github.com/dotnet/sdk/issues/433 -->


Youssef1313 (commented):
Can you update that in [dotnet/roslyn-sdk](https://github.com/dotnet/roslyn-sdk/blob/main/src/VisualStudio.Roslyn.SDK/Roslyn.SDK/ProjectTemplates/CSharp/Diagnostic/Vsix/Deployment.csproj) (and the parallel VB)?

Also the sample in dotnet/docs should be kept in sync.

Adding @sharwell as well. He has been helping review changes for this particular sample.

BillWagner:
@Youssef1313 

> Can you update that in dotnet/roslyn-sdk (and the parallel VB in roslyn-sdk)?

> Also the sample in dotnet/docs should be kept in sync.

I was going to open a new PR for those once we verified this is the solution. (Or you can, now that we see this works).

I'll merge this later today, unless @sharwell has comments on why we shouldn't.

sharwell (changes requested):
I do not believe this change is correct (it wasn't for dev16, not yet sure about dev17).

csharp/roslyn-sdk/Tutorials/MakeConst/MakeConst.Vsix/MakeConst.Vsix.csproj
```diff
@@ -19,7 +19,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.VSSDK.BuildTools" Version="15.1.192" PrivateAssets="all" />
```
sharwell, line 22: 📝 This package _defines_ `VSToolsPath` and places Microsoft.VsSDK.targets in the NuGet cache.

csharp/roslyn-sdk/Tutorials/MakeConst/MakeConst.Vsix/MakeConst.Vsix.csproj
```diff
@@ -19,7 +19,8 @@
...
+    <PackageReference Include="Microsoft.VisualStudio.SDK" Version="16.0.206" ExcludeAssets="runtime" />
```
sharwell, line 22: 📝 This package is not necessary. Excluding runtime assets is also not necessary.

csharp/roslyn-sdk/Tutorials/MakeConst/MakeConst.Vsix/MakeConst.Vsix.csproj
```diff
@@ -36,8 +37,6 @@
 
   <Import Sdk="Microsoft.NET.Sdk" Project="Sdk.targets" />
 
-  <Import Project="$(VSToolsPath)\VSSDK\Microsoft.VsSDK.targets" Condition="Exists('$(VSToolsPath)\VSSDK\Microsoft.VsSDK.targets')" />
```
sharwell, line 39: 📝 This is not an automatic import. Failure to include this line breaking packaging and deployment.

sharwell (changes requested):
Prior to this change, the build already used NuGet packages and did not expect the Visual Studio SDK to be installed. After this change, the packaging and deployment are broken. We should not merge this.

sharwell:
I haven't seen the specific build error that led to this change, but it's my hypothesis that the change you're looking for here is just the addition of this property group to the VSIX project:

```xml
<PropertyGroup Condition="'$(BuildingInsideVisualStudio)' != 'true'">
  <!-- This property disables extension deployment for command line builds; required for some CI -->
  <DeployExtension>False</DeployExtension>
</PropertyGroup>
```

jmarolf:
updated PR here: https://github.com/dotnet/samples/pull/4594

PR #4598, opened 2021-05-25
Notify Observers if Element changed with Index[]

ngoetti:
Are there any downsides to this? I needed to make this adjustment so that I could properly bind to elements in the Dictionary.

## Summary

Notify Observers if Element changed with Index[]


BillWagner (approved):
Thanks for making this change @ngoetti I like the addition. I've reviewed the changes, and I'll :shipit: now.

PR #4672, opened 2021-06-23
Add WF samples

gewarren:
I was able to obtain the zip file of WF samples from the download center even though they're not available for the public to download. Let's put the samples in this repo instead.

Note: I built the project that's showing snippets5000 warnings and it builds fine in VS 2022.


BillWagner (approved):
This LGTM.

Let's ping @adegeo  for why these aren't building in the Snippets 5000. I'm betting it's because the VS builds use the released version of VS 2019, and some new project elements are involved here.

adegeo:
@gewarren @BillWagner I'm going to guess that the snippets config file to swap it from `dotnet` to `msbuild via visual studio` wasn't added and so that's why it wasn't compiling correctly.

gewarren:
@adegeo Could you help to create the snippet config files for these WF samples?

PR #4673, opened 2021-06-23
Update ComWrappers impl with best practices.

AaronRobinsonMSFT:
## Summary

Update the `ComWrappers` sample following the recommended guidance from the interop team.

### DIFF
--- a/core/interop/comwrappers/IDispatch/ComWrappersImpl.cs
+++ b/core/interop/comwrappers/IDispatch/ComWrappersImpl.cs
@@ -13,31 +13,24 @@ internal unsafe class ComWrappersImpl : ComWrappers
     private static readonly ComInterfaceEntry* wrapperEntry;
 
     // This class only exposes IDispatch and the vtable is always the same.
-    // The below isn't the most efficient but it is reasonable for prototyping.
     // If additional interfaces want to be exposed, add them here.
     static ComWrappersImpl()
     {
-        GetIUnknownImpl(out IntPtr fpQueryInteface, out IntPtr fpAddRef, out IntPtr fpRelease);
+        GetIUnknownImpl(out IntPtr fpQueryInterface, out IntPtr fpAddRef, out IntPtr fpRelease);
 
-        var vtbl = new IDispatchVtbl()
-        {
-            IUnknownImpl = new IUnknownVtbl()
-            {
-                QueryInterface = fpQueryInteface,
-                AddRef = fpAddRef,
-                Release = fpRelease
-            },
-            GetTypeInfoCount = Marshal.GetFunctionPointerForDelegate(IDispatchVtbl.pGetTypeInfoCount),
-            GetTypeInfo = Marshal.GetFunctionPointerForDelegate(IDispatchVtbl.pGetTypeInfo),
-            GetIDsOfNames = Marshal.GetFunctionPointerForDelegate(IDispatchVtbl.pGetIDsOfNames),
-            Invoke = Marshal.GetFunctionPointerForDelegate(IDispatchVtbl.pInvoke)
-        };
-        var vtblRaw = RuntimeHelpers.AllocateTypeAssociatedMemory(typeof(IDispatchVtbl), sizeof(IDispatchVtbl));
-        Marshal.StructureToPtr(vtbl, vtblRaw, false);
+        // See IDispatch definition - https://docs.microsoft.com/windows/win32/api/oaidl/nn-oaidl-idispatch
+        var vtblRaw = (IntPtr*)RuntimeHelpers.AllocateTypeAssociatedMemory(typeof(IDispatch), IntPtr.Size * 7);
+        vtblRaw[0] = fpQueryInterface;
+        vtblRaw[1] = fpAddRef;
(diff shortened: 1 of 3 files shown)


core/interop/comwrappers/IDispatch/ComWrappersImpl.cs
```diff
@@ -13,31 +13,24 @@ internal unsafe class ComWrappersImpl : ComWrappers
...
+
+        // See IDispatch definition - https://docs.microsoft.com/windows/win32/api/oaidl/nn-oaidl-idispatch
+        var vtblRaw = (IntPtr*)RuntimeHelpers.AllocateTypeAssociatedMemory(typeof(IDispatch), IntPtr.Size * 7);
+        vtblRaw[0] = fpQueryInterface;
+        vtblRaw[1] = fpAddRef;
+        vtblRaw[2] = fpRelease;
+        vtblRaw[3] = (IntPtr)(delegate* unmanaged<IntPtr, int*, int>)&IDispatchVtbl.GetTypeInfoCount;
+        vtblRaw[4] = (IntPtr)(delegate* unmanaged<IntPtr, int, int, IntPtr*, int>)&IDispatchVtbl.GetTypeInfo;
+        vtblRaw[5] = (IntPtr)(delegate* unmanaged<IntPtr, Guid*, IntPtr*, int, int, int*, int>)&IDispatchVtbl.GetIDsOfNames;
+        vtblRaw[6] = (IntPtr)(delegate* unmanaged<IntPtr, int, Guid*, int, ComTypes.INVOKEKIND, ComTypes.DISPPARAMS*, IntPtr, IntPtr, IntPtr, int>)&IDispatchVtbl.Invoke;
```
jkoritzinsky, line 29: We could still use a typed vtable type with the function pointer types for the fields. That would make this slightly more readable.

core/interop/comwrappers/IDispatch/ComWrappersImpl.cs
```diff
@@ -13,31 +13,24 @@ internal unsafe class ComWrappersImpl : ComWrappers
...
-            GetIDsOfNames = Marshal.GetFunctionPointerForDelegate(IDispatchVtbl.pGetIDsOfNames),
-            Invoke = Marshal.GetFunctionPointerForDelegate(IDispatchVtbl.pInvoke)
-        };
-        var vtblRaw = RuntimeHelpers.AllocateTypeAssociatedMemory(typeof(IDispatchVtbl), sizeof(IDispatchVtbl));
-        Marshal.StructureToPtr(vtbl, vtblRaw, false);
-
-        wrapperEntry = (ComInterfaceEntry*)RuntimeHelpers.AllocateTypeAssociatedMemory(typeof(IDispatchVtbl), sizeof(ComInterfaceEntry));
-        wrapperEntry->IID = typeof(IDispatch).GUID;
-        wrapperEntry->Vtable = vtblRaw;
+        GetIUnknownImpl(out IntPtr fpQueryInterface, out IntPtr fpAddRef, out IntPtr fpRelease);
```
kant2002, line 19: I was suggested to use `GetIUnknownImpl(out vtblRaw[0], out vtblRaw[1], out vtblRaw[2]);` to reduce boilerplate previously in related discussions. Is this advice changed or here is inappropriate to show that to others?

core/interop/comwrappers/IDispatch/ComWrappersImpl.cs
AaronRobinsonMSFT: I don't think that actually adds much value. The type isn't needed in any way other development so if we are intending this sample to be the canonical way of how to implement this I would prefer to just implement it how a source generator should and avoid all superfluous metadata bloat.

core/interop/comwrappers/IDispatch/ComWrappersImpl.cs
AaronRobinsonMSFT: I can change it to that.

core/interop/comwrappers/IDispatch/ComWrappersImpl.cs
jkoritzinsky: Yeah the main value of it is removing all of the casts here, but it's understandable if we don't want to go that route.

core/interop/comwrappers/IDispatch/ComWrappersImpl.cs
AaronRobinsonMSFT: Actually after looking at that I think I am going to keep it as is. The main reason is if someone adds another interface to support calling `GetIUnknownImpl` isn't needed and so will need to be teased out to reuse the pointers.

core/interop/comwrappers/IDispatch/ComWrappersImpl.cs
AaronRobinsonMSFT: Agree the casting is annoying &ndash; I really dislike writing these manually.

jkotas (approved):

BillWagner (approved):
Thanks @AaronRobinsonMSFT 

I'll :shipit: now.

PR #4814, opened 2021-09-15
Update Excel Demo Readme

MSDN-WhiteKnight:
- Annotate for Sample Browser
- Add information how to build and run project from command line

.NET CLI does not support COM references, but the full MSBuild command line shipped with Visual Studio does. So it might be useful to have build instructions that don't depend on VS GUI.


BillWagner (approved):
Thanks @MSDN-WhiteKnight 

I've reviewed this, and I'll :shipit: now.  Thanks again.

PR #5032, opened 2022-03-01
Synch with the latest code from the docs

IEvangelist:
Synch with the latest code from the docs


BillWagner (approved):
This is a great update @IEvangelist 

It shows how much more clear the new syntax is.  Let's :shipit:

PR #5091, opened 2022-04-11
Azure App Config sample

CamSoper:
Azure App Config dynamic settings sample for .NET 6


azure/app-configuration-dynamic-reload/Models/PageSettings.cs
```diff
@@ -0,0 +1,9 @@
+namespace app_configuration_dynamic_reload.Models;
+
+public class PageSettings
```
BillWagner, line 3: Depending on the use, I would consider making this a `record`. That would create value based equality, and add a ToString, copy constructor, and with expression support.

BillWagner (approved):
This LGTM @CamSoper 

I had one optional suggestion. You can :shipit: when ready.

PR #5113, opened 2022-04-27
Add threadpool starvation endpoints

noahfalk:
We needed to demonstrate threadpool starvation for the tutorial I am adding.
https://github.com/dotnet/diagnostics/issues/2907


BillWagner (approved):
Thanks for adding these @noahfalk 

This LGTM, and I'll merge it now.

On note as you work on the related article:  Make sure you include text in the article that repeats the explanations in the comments.  Comments in code aren't localized on the docs platform, only the markdown text is translated. Non-English speakers should get all the information they need from the article text.

PR #5208, opened 2022-06-01
Less code, time & memory

snehil002:
Change of Logic - Without - for, if, modulo; however, gives correct answer
Applied little Maths: Arithmetic Progression
General formula for any sum between range 1 to limit, divisible by 'given' number
firstTerm needs to be guessed (here it is 3; because first term divisible by 3 in range 1 to 20 is 3)
commonDifference need to be > 0 ('given' divisor number is commonDifference)
limit is the upper limit of the given range (here 20)
Rest is Calculation in the Integer 32 bit range
Here AP is: 3, 6, 9, 12, 15, 18 = 63

## Summary

Describe your changes here.

### DIFF
--- a/csharp/branches-quickstart/Program.cs
+++ b/csharp/branches-quickstart/Program.cs
@@ -84,14 +84,8 @@ void ExploreLoops()
 
 void ChallengeAnswer()
 {
-    int sum = 0;
-    for (int number = 1; number < 21; number++)
-    {
-        if (number % 3 == 0)
-        {
-            sum = sum + number;
-        }
-    }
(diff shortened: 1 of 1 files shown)


BillWagner:
Hi @snehil002 

I do like your answer here. I am going to close this without merging it though. This sample is meant for beginners. It's focused on using the techniques they've just learned. While the code you've written is more efficient in all the metrics you've noted, it's less useful as a teaching aid.

snehil002:
Hi @BillWagner

I understood your point.
Thanks for acknowledging.

PR #5209, opened 2022-06-06
Update version-sweep.yml

IEvangelist:
## Summary

Fix version sweeper permissions, so that it can correctly access the resources.

Addresses: https://github.com/dotnet/samples/actions/runs/2418479162


BillWagner (approved):
LGTM. Let's :shipit:

PR #5236, opened 2022-07-25
Update Program.cs

davyzhu:
## Summary

recurring.AddMonths(2) won't change recurring's value.


BillWagner (approved):
Thanks for spotting this and making the change @davyzhu 

I've revied the changes and I'll :shipit: now.

PR #5627, opened 2023-01-07
Update to .NET 7 paths and add error reporting

kant2002:
I by mistake compile from x86 dev tools on Windows and that cause silent error failure. Hopefully my error reporting does not clutter code too much.

/cc @MichalStrehovsky


core/nativeaot/NativeLibrary/LoadLibrary.c
```diff
@@ -60,6 +60,16 @@ int callSumFunc(char *path, char *funcName, int firstInt, int secondInt)
     #else
         void *handle = dlopen(path, RTLD_LAZY);
     #endif
+    if (!handle)
```
MichalStrehovsky, line 63: Would it make sense to extract the whole library loading business into a separate function? We were already duplicating 5 lines before this change, but now it's 14 lines.

core/nativeaot/NativeLibrary/LoadLibrary.c
```diff
@@ -60,6 +60,16 @@ int callSumFunc(char *path, char *funcName, int firstInt, int secondInt)
...
+    {
+#ifdef _WIN32
+        int errorCode = GetLastError();
+        printf("Failed to load library at specified path. Error code: %d\n", errorCode);
+#else
+        puts("Failed to load library at specified path");
```
MichalStrehovsky, line 69: Nit: `dlerror` would be the equivalent of `GetLastError` on the Windows side.

MichalStrehovsky (commented):
Thanks!

kant2002:
@MichalStrehovsky can you review again?

gewarren (approved):

PR #5639, opened 2023-01-23
Add ref safety sample.

BillWagner:
This sample has the starter app and finished application for the tutorial added in dotnet/docs#32997

The tutorial shows the reader how to analyze when your application is allocating memory unnecessarily, and how you can use the profiler tools and the C# language `ref` safety features to improve performance.


CamSoper (approved):
I <3 this sample

PR #5765, opened 2023-03-28
Fix YAML in README.md

CamSoper:
Realized this sample (which is over a year old) was never properly indexed for the sample browser. Removed leading space in the YAML block to fix.


BillWagner (approved):
This LGTM @CamSoper 

Let's :shipit:

PR #5782, opened 2023-04-18
Fix failed samples

gewarren:
See https://github.com/MicrosoftDocs/samples/blob/c893868d5831b1ff4e66938a94a1bea548731725/dotnet/samples/.samples-status.json#L28-L41.

### DIFF
--- a/csharp/expression-trees/README.md
+++ b/csharp/expression-trees/README.md
@@ -2,36 +2,29 @@
 languages:
 - csharp
 products:
-- dotnet-csharp
+- dotnet
 page_type: sample
 name: "Build and interpret expression trees"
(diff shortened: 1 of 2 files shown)


csharp/expression-trees/README.md
```diff
@@ -2,36 +2,29 @@
...
 
 ```
-dotnet restore
 dotnet run
 ```
 
-`dotnet restore` restores the dependencies for this sample.
-
 `dotnet run` builds the sample and runs the output assembly.
-
```
gewarren, line 35: .NET Core 2.0 is so long ago that I don't think this note is needed anymore.

BillWagner (approved):
This LGTM @gewarren 

Let's :shipit:

Issue #5916, opened 2023-06-21
ArgumentOutOfRangeException thrown in Orleans Adventure Sample


ondrelogin:
**Error in Orleans Adventure Sample**

When running the `/orleans/Adventure` sample project, if you have more than one item in your inventory and you _end/die_ then a `System.ArgumentOutOfRangeException` will be thrown.

### Repro Steps
- Open the solution and run the server/Grain and the client.
- In the client console enter a name
- `go north`
- `go north` _again_
- `take brass key`
- `go west`
- `take knife`
- _optional_ : Enter `inv` to show that you have both the _brass key_ and _knife_  in your inventory
- Enter `end` to quit
- The Exception will be thrown

Here is some of the most relevant information from the exception that was thrown.
```
System.ArgumentOutOfRangeException
AdventureGrains.PlayerGrain.AdventureGrainInterfaces.IPlayerGrain.Die
  => AdentureGrains/PlayerGrain.cs:line 29
```

**Target framework**

Check the .NET target framework(s) being used, and include the version number(s).

.NET Core : _net7.0_

PR #5917, opened 2023-06-21
#5916 fix multiple drop things

ondrelogin:
## Summary

Creating a copy of the class level `_things` and iterating over the copy to _Drop_ the items instead of the actual original list.

### DIFF
--- a/orleans/Adventure/AdventureGrains/PlayerGrain.cs
+++ b/orleans/Adventure/AdventureGrains/PlayerGrain.cs
@@ -26,8 +26,13 @@ public class PlayerGrain : Grain, IPlayerGrain
     async Task IPlayerGrain.Die()
     {
         // Drop everything
-        var tasks = _things.Select(Drop).ToList();
-        await Task.WhenAll(tasks);
+        var dropTasks = new List<Task>();
+        var copy = new List<Thing>(_things);
+        foreach (var thing in copy)
+        {
+            dropTasks.Add(this.Drop(thing));
+        }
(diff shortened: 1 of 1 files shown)


orleans/Adventure/AdventureGrains/PlayerGrain.cs
```diff
@@ -26,8 +26,13 @@ public override Task OnActivateAsync(CancellationToken cancellationToken)
...
         // Drop everything
-        var tasks = _things.Select(Drop).ToList();
-        await Task.WhenAll(tasks);
+        var dropTasks = new List<Task>();
+        var copy = new List<Thing>(_things);
+        foreach (var thing in copy)
+        {
+            dropTasks.Add(this.Drop(thing));
+        }
+        await Task.WhenAll(dropTasks);
```
IEvangelist, line 35:
```suggestion
        var dropTasks = new List<Task<string?>>();        
        foreach (var thing in _things.ToArray() /* New collection */)
        {
            dropTasks.Add(Drop(thing));
        }
        await Task.WhenAll(dropTasks);
```

Let's do this instead.

IEvangelist (approved):

IEvangelist:
> ## Summary
> Creating a copy of the class level `_things` and iterating over the copy to _Drop_ the items instead of the actual original list.

I tried this out locally, you're correct! I like the changes, but wanted to stay closer to the original version. If we just call `.ToArray` it creates a new instance.

PR #6011, opened 2023-07-15
Fix parameter order

Grizzlly:
## Summary

The parameter order of the `AreEqual` method is `object expected, object actual`, thus the parameters have to be swapped.

[Docs](https://learn.microsoft.com/en-us/visualstudio/msbuild/tutorial-test-custom-task?view=vs-2022#unit-test) should be updated too.


BillWagner (approved):
This LGTM @Grizzlly 

I'll :shipit: now.

PR #6033, opened 2023-08-05
Replace mention of .NET Core

MSDN-WhiteKnight:
This sample now targets .NET 7, so it should not be called .NET Core application.


gewarren (approved):
Thanks @MSDN-WhiteKnight

PR #6150, opened 2023-09-12
Fix typo

UrtsiSantsi:
Use "component" instead of "compenent"


BillWagner (approved):
Thanks @UrtsiSantsi 

I'll :shipit: now.

PR #6153, opened 2023-09-13
Update WCF GettingStarted Sample to Conditionally Support .NET Framework and Core

afifi-ins:


mconnew (approved):

mconnew (changes requested):
Move the import to the end of the file

PR #6164, opened 2023-09-18
Upgrade CLI templates to latest article

adegeo:


adegeo:
This can't pass snippets5000 due to it being a template that contains other projects. We cannot write a bypass rule for it in the snippets5000 config file either because the errors are structural and not compiler-based.

BillWagner (approved):
This LGTM @adegeo 

Let's :shipit:

PR #6217, opened 2023-10-06
Update native AOT sample for .NET 8

jkotas:


core/nativeaot/NativeLibrary/README.md
```diff
@@ -86,16 +86,10 @@ The sample [source code](Class1.cs) demonstrates common techniques used to stay
...
 ```
 
-where `<Configuration>` is your project configuration (such as Debug or Release) and `<RID>` is the runtime identifier (one of win-x64, linux-x64, osx-x64). For example, if you want to publish a release configuration of your library for a 64-bit version of Windows the command would look like:
-
-```bash
-> dotnet publish /p:NativeLib=Static -r win-x64 -c release
-```
-
-The above command will drop a static library (Windows `.lib`, OSX/Linux `.a`) in `./bin/[configuration]/net7.0/[RID]/publish/` folder and will have the same name as the folder in which your source file is present.
+The above command will drop a static library (Windows `.lib`, macOS/Linux `.a`) in `./bin/[configuration]/net8.0/[RID]/publish/` folder and will have the same name as the folder in which your source file is present.
```
richlander, line 92: We could assume `Release` here instead of `[configuration]`.

core/nativeaot/HelloWorld/Dockerfile
```diff
@@ -8,9 +8,9 @@ RUN apt-get update \
 WORKDIR /source
 
 COPY . .
-RUN dotnet publish -c release -r linux-x64 -o /app
+RUN dotnet publish -o /app
```
richlander, line 11: Users will get a warning here if they open this project with Dev Kit because it will create a solution file. Best practice now is to add the project file to this line to avoid that.

richlander (approved):

BillWagner (approved):
This LGTM.

Issue #5640, opened 2023-01-24
.net code metrics failed after latest changes


pashkapashka:
**Before you open an issue**

If the issue is:

pipeline: https://github.com/dotnet/samples/actions/runs/3480443578
commit: https://github.com/dotnet/samples/commit/b74154f90fba167d2323d81fe07fefb0e554f799

`    - name: Set output
      run: |
        echo "updated-metrics=updatedMetrics" >> $GITHUB_OUTPUT
        echo "summary-title=title" >> $GITHUB_OUTPUT
        echo "summary-details=summary" >> $GITHUB_OUTPUT`

can't set output, have error:

Unable to process file command 'output' successfully.

<details>
<summary><strong>dotnet --info output</strong> or <strong>About VS info</strong></summary>

```console
<replace>
```
</details>

erichiller:
Same here, the issue is multiline output. Some more information [here](https://stackoverflow.com/questions/74137120/how-to-fix-or-avoid-error-unable-to-process-file-command-output-successfully).

This is the problem line:
https://github.com/dotnet/samples/blob/66506ef1d98d3579a35708717325ac6e5e676f28/github-actions/DotNet.GitHubAction/DotNet.GitHubAction/Program.cs#L57

IEvangelist:
Closed by #6272

PR #6272, opened 2023-10-21
GitHub actions summary output fix

erichiller:
This is a fix for #5640 to update the output format used by GitHub Actions.
Documentation here:
- https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
- https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-of-a-multiline-string


github-actions/DotNet.GitHubAction/DotNet.GitHubAction/Program.cs
```diff
@@ -79,7 +79,10 @@ static async Task StartAnalysisAsync(ActionInputs inputs, IHost host)
         {
             textWriter.WriteLine($"updated-metrics={updatedMetrics}");
             textWriter.WriteLine($"summary-title={title}");
-            textWriter.WriteLine($"summary-details={summary}");
+            textWriter.WriteLine(
+                "summary-details<<EOF\n" +
+                summary +
+                "\nEOF\n" );
```
adegeo, line 85: Just nit picking on style 😁 

I would think it would be better to do multiple write lines instead of a single with multiple lines, or use the interpolated string. This one is very clear as to the intent of the output:
```csharp
textWriter.WriteLine("summary-details<<EOF");
textWriter.WriteLine(summary);
textWriter.WriteLine("EOF");
```
or this, which is less clear but a one-liner:
```csharp
textWriter.WriteLine($"summary-details<<EOF\n{summary}\nEOF\n");
```
@IEvangelist ping

erichiller: Sure thing. Updated to use multiple `testWriter.WriteLine`. Let me know if there is anything else.

IEvangelist (approved):

PR #6281, opened 2023-10-31
Update WCF sample Instancing/Lifetime to conditionally support .NET framework / Core

afifi-ins:


mconnew:
Everything so far looks good. Need to update coding style to:
* Specify accessibility of members
* Use auto properties
* One line methods use expression method bodies
* Update private field naming to current convention of starting with _
* Make constant strings const

mconnew (approved):
:shipit:

PR #6282, opened 2023-10-31
Update WCF sample Compression Encoder code to conditionally support NET framework / Core

afifi-ins:


framework/wcf/Extensibility/MessageEncoder/Compression/CS/GZipEncoder/GZipMessageEncodingBindingElement.cs
```diff
@@ -20,68 +19,73 @@ static class GZipMessageEncodingPolicyConstants
     }
 
     //This is the binding element that, when plugged into a custom binding, will enable the GZip encoder
+#if NET462_OR_GREATER
     public sealed class GZipMessageEncodingBindingElement
```
mconnew, line 23: Can simplify with a single `#if` block
```c#
    public sealed class GZipMessageEncodingBindingElement 
                        : MessageEncodingBindingElement //BindingElement
#if NET462_OR_GREATER
                        , IPolicyExportExtension
#endif
```

PR #6563, opened 2024-01-11
Update README.md

xela-trawets:
include a reminder to avoid the 32 bit tooling on windows

## Summary

With Visual studio installed the default "developer command prompt" accesses the 32 bit versions of c compile and link (cl.exe). This results in a silent failure of the c sample code calling functions from the native aot library project.

Added the following line to the readme:

Please note that on Windows the platform (x86 and x64) must match between the dotnet aot library and the tool used to compile and link the c code; see [here](https://learn.microsoft.com/en-us/cpp/build/how-to-enable-a-64-bit-visual-cpp-toolset-on-the-command-line?view=msvc-170) for x64.


IEvangelist:
There were no changes to the README.md, was this intentional @xela-trawets?

xela-trawets:
Ah... let me try that again.. sorry

xela-trawets:
closing this (not enough of a git-ninja to fix it), 
moved to a fresh patch
https://github.com/dotnet/samples/pull/6665

(not entirely sure what a git-ninja would be, but I am not that)

PR #6651, opened 2024-02-13
Add a warning to building static libraries with naot

MichalStrehovsky:
Cc @jkotas @lambdageek 

Suggestions for a better wording welcome!

Fixes dotnet/runtime#98356.

### DIFF
--- a/core/nativeaot/NativeLibrary/README.md
+++ b/core/nativeaot/NativeLibrary/README.md
@@ -9,7 +9,7 @@ Create a .NET class library project using `dotnet new classlib -o NativeLibrary`
 ## Building shared libraries
 
 ```bash
-> dotnet publish /p:NativeLib=Shared --use-current-runtime
+> dotnet publish --use-current-runtime
 ```
 
 The above command will drop a shared library (Windows `.dll`, macOS `.dylib`, Linux `.so`) in `./bin/Release/net8.0/[RID]/publish/` folder and will have the same name as the folder in which your source file is present.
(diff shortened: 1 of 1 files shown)


core/nativeaot/NativeLibrary/README.md
```diff
@@ -85,6 +85,9 @@ The sample [source code](Class1.cs) demonstrates common techniques used to stay
 
 ## Building static libraries
 
+> [!WARNING]
+> Building shared libraries is preferred to building static libraries. Static libraries might have problems with ABI differences (the compiler-rt runtime used to build the static libraries that ship with native AOT needs to match the compiler-rt runtime used by other code in the same loadable module). It is also not not possible to mix multiple native AOT compiled static libraries within the same loadable module. These problems don't exist when building a shared library.
```
jkotas, line 89:
```suggestion
> Building shared libraries is preferred to building static libraries. All code in the loadable module must be compiled with C/C++ compiler version and options that are compatible with native AOT static libraries. It is also not possible to mix multiple native AOT compiled static libraries within the same loadable module. These problems don't exist when building a shared library.
```
`compiler-rt` is one specific example. I would say it in more general way.

gewarren, line 89:
```suggestion
> It's preferred to build shared libraries than static libraries:
>
> - Static libraries might have problems with ABI differences&mdash;the compiler-rt runtime used to build the static libraries that ship with Native AOT must match the compiler-rt runtime used by other code in the same loadable module.
> - It's also not possible to mix multiple Native AOT compiled static libraries within the same loadable module.
>
> These problems don't exist when you build a shared library.
```

gewarren (approved):
I left some suggestions to consider.

MichalStrehovsky:
Manually incorporated feedback from both @jkotas and @gewarren since there were conflicting lines.

PR #6745, opened 2024-03-15
Updated Orleans Chipter sample to net8.0

m3nax:
## Summary

Updated Orleans Chipter sample to net8.0


BillWagner (approved):
Thanks again. This LGTM.

PR #6746, opened 2024-03-15
Updated Orleans FSharp sample to net8.0

m3nax:
## Summary

Updated Orleans FSharp sample to net8.0


BillWagner (approved):
This LGTM. I'll :shipit: now.

PR #6747, opened 2024-03-15
Updated Orleans GPSTracker example to net8.0

m3nax:
## Summary

Updated Orleans GPSTracker example to net8.0


BillWagner (approved):
This LGTM. I'll :shipit:

PR #6751, opened 2024-03-18
Updated Orleans Presence example to net8.0

m3nax:
## Summary

Updated Orleans Presence example to net8.0


BillWagner (approved):
This LGTM. Let's :shipit:

PR #6757, opened 2024-03-26
Updated NUnit testing samples for NUnit 4

rextor92:
## Summary

- I've updated the C#, VB and F# samples that were using NUnit to address the breaking changes in NUnit4.
- All projects in the folder updated to .NET 8
- Misc: .NET Standard version updated for one project, error for duplicate reference to MSTest.TestAdapter resolved
- Ran all projects to validate, all tests passing

Fixes dotnet/docs#40080 
Tagging @IEvangelist for review

### DIFF
--- a/core/getting-started/unit-testing-with-fsharp-nunit/MathService.Tests/Tests.fs
+++ b/core/getting-started/unit-testing-with-fsharp-nunit/MathService.Tests/Tests.fs
@@ -6,14 +6,6 @@ open NUnit.Framework
 open MathService
 
 [<Test>]
-let ``My test`` () =
-    Assert.True(true)
-
-//[<Test>]
-//let ``Fail every time`` () = Assert.True(false)
-
-
-[<Test>]
 let ``Sequence of Evens returns empty collection`` () =
     let expected = Seq.empty<int>
     let actual = MyMath.squaresOfOdds [2; 4; 6; 8; 10]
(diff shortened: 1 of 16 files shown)


core/getting-started/unit-testing-with-fsharp/MathService.Tests/Tests.fs
```diff
@@ -5,14 +5,6 @@ open Xunit
 
 open MathService
 
-[<Fact>]
```
rextor92, line 8: I understand why this existed, but with the NUnit changes it no longer makes sense so I completely removed it @IEvangelist

rextor92:
@dotnet-policy-service agree

core/getting-started/unit-testing-with-fsharp-nunit/MathService.Tests/Tests.fs
```diff
@@ -5,14 +5,6 @@ open NUnit.Framework
 
 open MathService
 
-[<Test>]
```
rextor92, line 8: I understand why this existed, but with the NUnit changes it no longer makes sense so I completely removed it. Wanted to explicitly point it out as it looks like "no test, no problem".

BillWagner (approved):
This LGTM @rextor92 

I'll :shipit: now.

PR #6939, opened 2024-06-05
Add comments about .NET unload

agocke:
## Summary

Note in the .NET hosting docs that the runtime cannot be unloaded.


agocke:
@dotnet/appmodel for review

PR #6988, opened 2024-09-16
Fix Orleans ChatRoom sample to match documentation

reagcz:
## Summary

Change the ChatRoom client sample to use the `streamId` returned by the `IChannelGrain.Join` and `Leave` methods

The [README.md](https://github.com/dotnet/samples/blob/4231a563d5fecdcfaebe8de6d43cef0af859ec90/orleans/ChatRoom/README.md?plain=1#L25) for the ChatRoom sample states that

> Clients connect to the `ChannelGrain` for a channel and then subscribe to the stream identified by the `Guid` returned from the `IChannelGrain.Join` call

The sample currently creates the streamId manually, instead of using the one returned by the Join and Leave methods.


BillWagner (approved):
This LGTM, and I'll :shipit: now.

PR #6997, opened 2024-11-08
Add documentation for building native libraries with NativeAOT targeting iOS-like platforms

ivanpovazan:
## Summary

This PR adds documentation describing how to build and consume native libraries with NativeAOT targeting iOS-like platforms.

---

Fixes https://github.com/dotnet/runtime/issues/96742
Contributes to: https://github.com/dotnet/runtime/issues/96664


core/nativeaot/NativeLibrary/ios-like-native-libraries-with-nativeaot.md
```diff
@@ -0,0 +1,250 @@
+# iOS-like Native Libraries with NativeAOT in .NET 9
```
jkotas, line 1: This is samples repo, so there should a working sample that shows you how to do it.

If you do not want to turn this into an actual sample, should this rather be in the docs?

ivanpovazan: Yes, you are right, this fits better in the docs repository.
I created a PR there instead.

ivanpovazan:
Closing in favour of https://github.com/dotnet/docs/pull/43397

PR #7009, opened 2025-01-18
Ensure CustomDataAdapter works in cluster with multiple silos

egil:
## Summary

The sample requires `[GenerateSerializer]` and friends to be added to `CustomBatchContainer` to work in a cluster with multiple silos. Without it a `CodecNotFoundException` exception is thrown.

It is not clear to me if the property `StreamId` is also serialized, and if it needs to, it should probably also have the `[Id]` attribute.


BillWagner:
ping @IEvangelist

IEvangelist (approved):

PR #7012, opened 2025-01-19
Upgrade orleans blazor sample

egil:
This upgrades the orleans blazor server and blazor wasm samples.

Note, the Blazor projects themselves were based on an older template, so I recreated them using the Blazor templates from dotnet 9. That is why there are so many changes in this PR.


BillWagner:
ping @IEvangelist

IEvangelist (approved):

PR #7019, opened 2025-02-21
Bug fixes for Orleans Bank Account sample.

Rossimac:
Balance now updates after Deposit or Withdraw operations.

Withdraw decrements now instead of incorrectly incrementing.


oising:
Removing the immutability of the Balance class sets a bad precedent for this sample. The idea of immutability is to prevent the class of bugs where grain A makes a change to a instance, then sends it to grain B -- on the same silo -- and then makes another change to the instance, which, because it's a reference/byref, will change the instance in grain B also.

Rossimac:
I fixed a couple of build errors in the commit. 

Also, I fixed the sample to allow the bank balance to actually persist the balance after a Withdrawal.

PR #7029, opened 2025-03-14
Orleans JournaledGrain sample

egil:
This sample aims to showcase different ways to use JournaledGrains.

### DIFF
--- a/orleans/JournaledTodoList/JournaledTodoList.WebApp/Grains/TodoListRegistryGrain.cs
+++ b/orleans/JournaledTodoList/JournaledTodoList.WebApp/Grains/TodoListRegistryGrain.cs
@@ -1,0 +1,66 @@
+﻿using System.Collections.Immutable;
+using Orleans.Utilities;
+
+namespace JournaledTodoList.WebApp.Grains;
+
+public sealed class TodoListRegistryGrain(
+    [PersistentState("todoListRegistry", "Default")]
+    IPersistentState<TodoListRegistryGrain.TodoListRegistry> state,
+    ILogger<TodoListRegistryGrain> logger) : Grain, ITodoListRegistryGrain
+{
+    private readonly ObserverManager<ITodoListRegistryObserver> observers = new(
+        TimeSpan.FromMinutes(5),
+        logger);
+
+    public async Task RegisterTodoListAsync(string todoListId)
+    {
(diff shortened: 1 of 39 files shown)


orleans/JournaledTodoList/JournaledTodoList.WebApp/Grains/TodoListRegistryGrain.cs
```diff
@@ -0,0 +1,82 @@
+﻿using System.Collections.Immutable;
+using Orleans.EventSourcing;
+using Orleans.Providers;
+using Orleans.Utilities;
+
+namespace JournaledTodoList.WebApp.Grains;
+
+/// <summary>
+/// A "state" based Journaled Grain. It does not save the events, just the state.
+/// </summary>
```
egil, line 10: I would like to add something here that tells Orleans to have one copy of this grain on each silo. Using `[StatelessWorker(1)]` didn't seem to work, or at least, it resulted in notificaitons/subscriptions not working as expected.

egil: Nevermind, I was looking at the [Replicated grains](https://learn.microsoft.com/en-us/dotnet/orleans/grains/event-sourcing/replicated-instances) features that are no longer available.

orleans/JournaledTodoList/JournaledTodoList.WebApp/Program.cs
```diff
@@ -0,0 +1,39 @@
...
+var builder = WebApplication.CreateBuilder(args);
+
+builder.AddServiceDefaults();
+builder.AddKeyedAzureTableClient("clustering");
+builder.AddKeyedAzureBlobClient("grain-state");
+builder.UseOrleans(siloBuilder =>
+{
+    siloBuilder.AddLogStorageBasedLogConsistencyProviderAsDefault();
+    siloBuilder.AddStateStorageBasedLogConsistencyProvider(name: Constants.StateStorageProviderName);
+});
```
egil, line 13: With this set up, the log consistency providers are using the grain state as storage, in this case, blob storage. When the log-based providers save state, does it just save the added/raised event(s), or does it save all in one go? That is, if there are 1000 events, will they all be serialized and save to blob storage or just the new events?

ReubenBond (approved):

PR #7032, opened 2025-03-18
Update dotnet-code-metrics.yml

danmoseley:
use forked version of GHA to reduce security surface area. similar to https://github.com/dotnet/aspire/pull/8122/files


BillWagner (approved):
Thanks @danmoseley 

This LGTM.  I'll :shipit: now.

PR #7054, opened 2025-05-18
Upgrade DiagnosticScenarios to .NET 9

verdie-g:
## Summary

Upgrading DiagnosticScenarios to .NET 9 to be able use a new runtime event documented here https://github.com/dotnet/docs/pull/40852.


verdie-g:
@noahfalk could you have a quick look at this 🙏 It relates to https://github.com/dotnet/docs/pull/46251. The doc now mentions that .NET 9 is needed but this app uses .NET 8.

PR #7055, opened 2025-06-01
Fix broken link

ericmutta:
## Summary

Original document was linking to:

> https://github.com/dotnet/samples/blob/main/core/nativeaot/NativeLibrary/Class1.cs

...which doesn't exist. I reckon the intent was to link to this instead:

> https://github.com/dotnet/samples/blob/main/core/nativeaot/NativeLibrary/LibraryFunctions.cs


BillWagner (approved):
Thanks @ericmutta 

This LGTM, and I'll :shipit:

ericmutta:
Thanks for the prompt followup @BillWagner 🚀

PR #7056, opened 2025-06-07
Update AccountGrain.cs

### LINKED ISSUE #7021: Orleans Bank Account/Transaction sample is broken
## [Change Title]

- The sample code does not update the (override the old) grain state. 
- As a result, the balance of all accounts remains 1000 even though they are transacting.  

---
#### Issue metadata

* Issue type: sample-update

krsion:
Reverted usage of immutable balance, fixed Orleans Bank Account Transaction Sample.
Fixes #7021

### DIFF
--- a/orleans/BankAccount/AccountTransfer.Grains/AccountGrain.cs
+++ b/orleans/BankAccount/AccountTransfer.Grains/AccountGrain.cs
@@ -1,13 +1,13 @@
 ﻿using AccountTransfer.Interfaces;
 using Orleans.Concurrency;
-using Orleans.Transactions.Abstractions;
+//using Orleans.Transactions.Abstractions;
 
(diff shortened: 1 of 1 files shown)


Rossimac:
I have an open PR [here](https://github.com/dotnet/samples/pull/7019) and I've fixed up the build issues today. Also, I changed a few things about the AccountGrain and the Balance class so that any changes are persisted after a Withdrawal. I ran into the same issues you're seeing with the balances not actually changing after a transaction :)

ReubenBond:
I (belatedly) merged @Rossimac's existing PR, so I'll close this PR since there are no additional changes necessary. Thank you for the PR, @krsion - please don't hesitate to open more to fix other issues you find

PR #7092, opened 2026-02-13
Make bank system more flexible

ivanPeshterskii:
By creating some interfaces, the project gets more flexibility and it can be used more easily and changed by developers.


adegeo:
Hello. Thank you for your interest in improving the samples. These particular samples are supported documentation tutorials for new users. Introducing new concepts, such as interfaces, are outside the scope of that specific article.

I'm going to close this PR.

Tagging @BillWagner to let me know if I'm wrong in my thoughts here.

PR #7122, opened 2026-06-23
Fix race condition in Orleans shopping cart inventory

### LINKED ISSUE #7121: Race condition in orleans/ShoppingCart/Grains/InventoryGrain.cs#L48
A parallel for loop writes to a dictionary without mutual exclusion: https://github.com/dotnet/samples/blob/main/orleans/ShoppingCart/Grains/InventoryGrain.cs#L48

TimonPasslick:
## Summary

A dictionary gets filled in a parallel for loop without protection, so I added locking.

Fixes #7121

### DIFF
--- a/orleans/ShoppingCart/Grains/InventoryGrain.cs
+++ b/orleans/ShoppingCart/Grains/InventoryGrain.cs
@@ -40,7 +40,7 @@ public sealed class InventoryGrain(
             return;
         }
 
-        object lockObject = new();
+        Lock lockObject = new();
         await Parallel.ForEachAsync(
             state.State,
             async (id, _) =>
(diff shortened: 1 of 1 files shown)


orleans/ShoppingCart/Grains/InventoryGrain.cs
```diff
@@ -40,12 +40,17 @@ private async Task PopulateProductCacheAsync()
             return;
         }
 
+        object lockObject = new();
```
BillWagner: I'd prefer seeing the lock object be a `System.Threading.Lock` type.

TimonPasslick: Cool, I didn't know about this explicit lock type. Fixed

PR #7125, opened 2026-06-24
Fix state inconsistency in Orleans Chirper account

### LINKED ISSUE #7124: State inconsistency bug in orleans/Chirper/Chirper.Grains/ChirperAccount.cs
https://github.com/dotnet/samples/blob/main/orleans/Chirper/Chirper.Grains/ChirperAccount.cs

`FollowUserIdAsync‎` and `UnfollowUserIdAsync` may not interleave, but the grain is reentrant. Interleaving can cause 2 problems if both methods get called with the same username on the same grain:
1. If a user gets followed, but between the remote and local update, he gets unfollowed, the local state says the user is followed, but the remote state says the user is not followed. Same vice versa.
2. The follow and unfollow commands may arrive on the other grain in reverse of sending order, causing the same inconsistencies.

I don't see a way to keep the grain fully reentrant, which is unfortunate, because this is supposed to be a demonstration for reentrant grains: https://learn.microsoft.com/en-us/samples/dotnet/samples/orleans-chirper-social-media-sample-app/

TimonPasslick:
## Summary

Disabled reentrancy/interleaving for `FollowUserIdAsync`‎ and `UnfollowUserIdAsync`. Interleaving can cause 2 problems if both methods get called with the same username on the same grain:
1. If a user gets followed, but between the remote and local update, he gets unfollowed, the local state says the user is followed, but the remote state says the user is not followed. Same vice versa.
2. The follow and unfollow commands may arrive on the other grain in reverse of sending order, causing the same inconsistencies.

Fixes #7124

### DIFF
--- a/orleans/Chirper/Chirper.Grains/ChirperAccount.cs
+++ b/orleans/Chirper/Chirper.Grains/ChirperAccount.cs
@@ -3,10 +3,11 @@ using Chirper.Grains.Models;
 using Microsoft.Extensions.Logging;
 using Orleans.Concurrency;
 using Orleans.Runtime;
+using Orleans.Serialization.Invocation;
 
 namespace Chirper.Grains;
 
-[Reentrant]
+[MayInterleave(nameof(WontFollowAndUnfollowSimultaneously))]
 public sealed class ChirperAccount : Grain, IChirperAccount
 {
     /// <summary>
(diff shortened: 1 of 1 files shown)


TimonPasslick:
@dotnet-policy-service agree

